PR #6, opened 2021-06-30
Fix EVR page freeze issue

saba-ja:
| | |
|:---|:---|
|**_Originating Project/Creator_**| F Prime |
|**_Affected Component_**| GDS |
|**_Affected Architectures(s)_**| - |
|**_Related Issue(s)_**| - |
|**_Has Unit Tests (y/n)_**| No |
|**_Builds Without Errors (y/n)_**| Yes |
|**_Unit Tests Pass (y/n)_**| NA |
|**_Documentation Included (y/n)_**| Yes |

---
## Change Description

Current GDS Event page freezes when the number of EVRs in the event table reaches a certain size.
This issue makes the entire webpage irresponsive. 

## Rationale

The issue is DOM rendering is computationally expensive and having too many HTML elements in the DOM can effect browser's performance. To fix the issue this PR adds lazy loading of EVRs. Events are only added up to a certain number. After the maximum size user has to scroll down to the bottom of the page to load the next group of EVRs. Upon adding the new group the previous EVRs are removed from the DOM.

Manually tested and verified root cause of the slow down is the number of HTML elements and not memory issues.
[GDS with issue](https://youtu.be/lOc2yhHsg1M)

Manually verified GDS functionality after the fix. As shown in the video below the EVR table had ~28K EVRs and the page did not show any sign of slow down:
[GDS EVR fixed](https://youtu.be/Oeu9BapAxnE)

Functionalities:

- Load next EVR group after scrolling down.
- Load prev EVR group after scrolling up.
- Can use `First` button to move to the top of the list.
- Can use `Last` button to move to the bottom of the list.
- Can stay at the bottom when pressing `Last` button or scrolling down to the last element of the list.
- Will activate auto update when at the end of the list. Shows a green flag to indicate auto update is enabled.
- Shows total size of EVR list.
- Show start and end of current EVR window.

## Future Work

NA


LeStarch (approved):
The video really helped!

PR #9, opened 2021-07-23
lestarch: command raw text input and fixed client validation

LeStarch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| OWLS |
|**_Affected Component_**| Validation |
|**_Related Issue(s)_**|  |

---
## Change Description

Adds in much better validation, error handling, and error reporting.  Provides a raw command input that syncs with the selected command input for copy/past in and out of commands.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -57,6 +57,7 @@ bytearray
 calibri
 callergraph
 callgraph
+cargs
 cata
 cbl
 CCB
@@ -200,6 +201,7 @@ figcaption
 filemode
 fileno
 filepath
+fileurl
 finditer
 Firefox
 flac
@@ -293,6 +295,7 @@ janamian
 javadoc
 javascript
 Jax
+Jdk
 joinpath
 jpe
 jpeg
@@ -376,6 +379,7 @@ noapp
 noqa
 nosort
 Noto
+novalidate
 nowait
 nowrap
 NSPACES
@@ -556,6 +560,7 @@ telem
 testcase
 TESTLIST
 textarea
+textbox
 tgz
 thead
 thtcp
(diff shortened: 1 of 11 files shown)


metzdigital:
It would be nice if it "auto-capitalize" text submitted:
image

metzdigital:
A bug found, it appears `0` is not seen as a valid integer argument:
image

metzdigital:
I'm not sure how much this matters, but I noticed there is no "auto-casing" for the command, just the arguments:
image

metzdigital:
An interesting behavior I noticed. For the ADC_ADS1119_TESTING command arg1 thru arg3 are labeled as U32:
```
<command kind="guarded" opcode="0" mnemonic="ADC_ADS1119_TESTING">
	<comment>
		Command to run tests against ADS1119 driver
	</comment>
	<args>
		<arg type="Backplane::BpHwTestNames" name="test"> </arg>
		<arg type="U32" name="arg1"> </arg>
		<arg type="U32" name="arg2"> </arg>
		<arg type="U32" name="arg3"> </arg>
		<arg type="F32" name="arg4"> </arg>
	</args>
</command>
```
But values larger than 4294967295 still get the "green check":
image
Until I attempt to send the command:
image

metzdigital:
This isn't the most descriptive error, the real issue is too long of a string:
image

metzdigital:
The above items is all I could find when doing extensive hammering on it. Really well done @LeStarch!

LeStarch:
@metzdigital the zero not an int error is a true error, and I will fix it.  I look into the out-of-range error with the other....I can probably set limits, maybe, probably.

metzdigital:
Verified above items have been fixed

metzdigital (approved):
Approved.

PR #16, opened 2021-08-06
Ci workflow

aidan-wagner:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Added fprime-gds tests to workflow

## Rationale

To ensure fprime-gds is not broken by future updates

## Future Work

Create tests to ensure backwards compatability

### DIFF
--- a/.github/workflows/fprime-gds-tests.yml
+++ b/.github/workflows/fprime-gds-tests.yml
@@ -1,6 +1,6 @@
 name: fprime-gds tests
 
-on: [pull-request]
+on: [pull_request]
 
 jobs:
   build:
--- a/.github/workflows/fprime-gds-tests.yml
+++ b/.github/workflows/fprime-gds-tests.yml
@@ -1,6 +1,6 @@
 name: fprime-gds tests
 
-on: [pull_request]
+on: [pull_request, push]
 
 jobs:
   Unit-tests:
--- a/.github/workflows/fprime-gds-tests.yml
+++ b/.github/workflows/fprime-gds-tests.yml
@@ -1,6 +1,6 @@
 name: fprime-gds tests
 
-on: [push]
+on: [pull-request]
 
 jobs:
   build:
--- a/.github/workflows/fprime-gds-tests.yml
+++ b/.github/workflows/fprime-gds-tests.yml
@@ -3,7 +3,7 @@ name: fprime-gds tests
 on: [pull_request]
 
 jobs:
-  build:
+  Unit-tests:
 
     runs-on: ubuntu-latest
     strategy:
--- a/.github/workflows/fprime-gds-tests.yml
+++ b/.github/workflows/fprime-gds-tests.yml
@@ -1,6 +1,6 @@
 name: fprime-gds tests
 
-on: [pull_request]
+on: [push]
 
 jobs:
   build:
--- a/.github/workflows/fprime-gds-tests.yml
+++ b/.github/workflows/fprime-gds-tests.yml
@@ -6,13 +6,16 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/fprime-gds-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install dependencies
+        run: |
+          pip install -e .
+          python -m pip install --upgrade pip
+          pip install pytest
```
Joshua-Anderson: Is there a reason that you're manually installing pytest? It's a package dependency so it should automatically be installed with the rest of the GDS

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
```
Joshua-Anderson, line 1: Since we don't have multiple flavors of GDS unit tests, are markers necessary?

.github/workflows/fprime-gds-tests.yml
```diff
@@ -0,0 +1,23 @@
+name: fprime-gds tests
+
+on: [push]
```
Joshua-Anderson: I think you may need to add `pull_request` as well to run CI on pull requests

.github/workflows/fprime-gds-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
```
Joshua-Anderson: We should actually test a few different versions of python to make sure the GDS works in most python environments. I think we should test against python 3.6, the oldest version of python we support, and python 3.9, the newest python release.

You can use github action build matricies to easily run the same CI workflow a few times with different versions of python

pytest.ini
aidan-wagner: Ok, I'll remove the markers from the tests

test/fprime_gds/common/gds_cli/filtering_utils_test.py
```diff
@@ -54,7 +54,6 @@ def sample_event(sample_template):
         filtering_utils.get_search_predicate("the same combination"),
     ],
 )
-@pytest.mark.gds_cli
```
Joshua-Anderson: Do we need to remove these markers to get everything working? I haven't used pytest much, so I definitely don't know what I'm talking about, but I assumed that pytest would just run everything by default. If tests have already been marked they're probably worth keeping just on the off chance somebody running locally just wants to run the cli tests or something.

aidan-wagner: Yeah, if you have the markers but don't define them then it gives a bunch of warnings. Should I add them back in then?

Joshua-Anderson: Ah, I get it now. Yeah, add the pytest.ini file back in. Something tells me nobody has been running these unit tests anytime recently.

Joshua-Anderson (approved):
:shipit: Sorry for the churn on the pytest markers... looks good to me now!

.github/workflows/fprime-gds-tests.yml
```diff
@@ -0,0 +1,25 @@
+name: fprime-gds tests
+
+on: [pull_request]
```
Joshua-Anderson: Can you please make this workflow triggered on both pull requests and pushes (so that it runs on the devel branch as well)

Joshua-Anderson (changes requested):

LeStarch:
@aidan-wagner what tests does this run?  Do we have pytests in GDS that aren't integration tests?

LeStarch:
@aidan-wagner i stand corrected.  Looks good!

Joshua-Anderson (approved):

PR #50, opened 2022-04-08
Codebase formatting with ``f-string``

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---

## Change Description

- Replace calls to ``string.format()`` with ``f-strings``
- Replace the use of the ``%`` string interpolation operator with ``f-strings`` (only ``%s`` operator for the moment)

## Rationale

According to the [Python documentation](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting), using the ``%`` operator to format strings can lead to "a variety of quirks that lead to a number of common errors."

[PEP 498](https://peps.python.org/pep-0498/) added F-strings to Python in version 3.6. F-strings are a versatile and powerful method of string formatting. Because the code looks more like the output, they make it shorter and more readable.

Take a deep breath

## Future Work

Deals with operators other than ``%s`` like ``%d``.

### DIFF
--- a/src/fprime_gds/executables/run_deployment.py
+++ b/src/fprime_gds/executables/run_deployment.py
@@ -49,7 +49,7 @@ def find_app(root: Path) -> Path:
             files.append(child)
 
     if len(files) == 0:
-        print(f"[ERROR] binary location {bin_dir} does not exist")
+        print(f"[ERROR] App not found in {bin_dir}")
         sys.exit(-1)
 
     if len(files) > 1:
--- a/src/fprime_gds/executables/run_deployment.py
+++ b/src/fprime_gds/executables/run_deployment.py
@@ -31,9 +31,7 @@ def get_artifacts_root() -> Path:
         sys.exit(-1)
     assert "install_dest" in ini_settings, "install_dest not in settings.ini"
     print(
-        "[INFO] Autodetected artifacts root '{}' from deployment settings.ini file.".format(
-            ini_settings["install_dest"]
-        )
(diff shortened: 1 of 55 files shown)


ThibFrgsGmz:
@LeStarch Should I have split this PR into several to make code review easier?
In retrospect, I realize it's a lot of editing, but since it's mostly string formatting...

LeStarch:
@ThibFrgsGmz Yes, we absolutely need to start making smaller PRs that are easier to review.  I am guilty of this too, as I tend to make fairly involved PRs.

Since this is done, and string edits....I'll just go through it as is.  It will just take more time to review.

ThibFrgsGmz:
@LeStarch OK great, sorry for the burden...
The size of my current pending PRs on ``fprime-tools`` are better, right?

LeStarch:
Yes, much more manageable.

src/fprime_gds/executables/run_deployment.py
```diff
@@ -51,14 +49,12 @@ def find_app(root: Path) -> Path:
             files.append(child)
 
     if len(files) == 0:
-        print("[ERROR] No app found in binary location {}".format(bin_dir))
+        print(f"[ERROR] binary location {bin_dir} does not exist")
```
LeStarch: This error message changed.

LeStarch (approved):
Fixed one error message that changed.

src/fprime_gds/executables/run_deployment.py
LeStarch: I went ahead and fixed this one.

PR #69, opened 2022-06-03
lestarch: loosening package requirements

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

See: fprime-community/fprime-tools#78


Joshua-Anderson (commented):
Most python packages specify a major version of the packages in use ([apache airflow](https://github.com/apache/airflow/blob/main/setup.cfg#L83)). The goal is to prevent users from using incompatible package versions (i.e. super old versions of pytest) without overly constraining users from upgrading the dependencies. I'd recommend going with that approach if possible.

timcanham (dismissed):

timcanham (approved):

PR #74, opened 2022-06-17
Remove redundant slice index from ``range()``

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims to remove unnecessary explicit ``0`` argument from ``range()``.

## Rationale

It is not necessary to explicitly define the starting value of a ``range()`` call because the default value is ``0``.
This refactoring removes the zeros, which shortens the code slightly.

void

## Future Work

void

### DIFF
--- a/src/fprime_gds/common/communication/framing.py
+++ b/src/fprime_gds/common/communication/framing.py
@@ -232,7 +232,7 @@ class TcpServerFramerDeframer(FramerDeframer):
         if not no_copy:
             data = copy.copy(data)
         # Shift over to ZZZZ
-        while len(data) > 4 and data[4] != b"ZZZZ":
+        while len(data) > 4 and data[:4] != b"ZZZZ":
             data = data[1:]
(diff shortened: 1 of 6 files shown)


src/fprime_gds/common/communication/framing.py
```diff
@@ -232,7 +232,7 @@ def deframe(self, data, no_copy=False):
         if not no_copy:
             data = copy.copy(data)
         # Shift over to ZZZZ
-        while len(data) > 4 and data[0:4] != b"ZZZZ":
```
LeStarch, line 235: This still must read `[:4]`

LeStarch: I fixed it.

LeStarch (approved):

LeStarch (dismissed):

LeStarch (approved):

PR #77, opened 2022-06-17
[Deleted] Refactoring with list-comprehension

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims to convert some ``for`` loops into a list comprehension.

## Rationale

Comprehensions list allows to create the ``list`` on a single line, which avoids declaring an empty ``list`` and then adding elements.
Comprehensions generally run faster than building the collection in a loop, which is another factor if performance is an important consideration.

For the two ``*history_unit_test.py`` files, these comprehension create a copy of the original collection. They can all be simplified by simply building a new collection directly. The resulting code is even easier to read and shows the intent more clearly.

## Future Work

void

### DIFF
--- a/src/fprime_gds/executables/run_deployment.py
+++ b/src/fprime_gds/executables/run_deployment.py
@@ -43,12 +43,8 @@ def find_app(root: Path) -> Path:
         print(f"[ERROR] binary location {bin_dir} does not exist")
         sys.exit(-1)
 
-    files = []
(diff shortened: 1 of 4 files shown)


ThibFrgsGmz:
Oops, I didn't fetch and merge the last commits in my fork...

PR #82, opened 2022-06-29
Switch from `range(len(...))` to `enumerate(...)`

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims to switch from `range(len(...))` to `enumerate(...)` to access the counter and the value from the iterables at the same time.

## Rationale

It is not Pythonic to use `range(len(...))`. Index based loops do not exist in Python, it uses collection iterators instead.

Python includes an `enumerate` method to add a counter to an iterator. This method allows you to access both the counter and the value of the iterator at the same time. It is therefore recommended to replace `range(len(...))` with `enumerate(...)`.

void

## Future Work

void


LeStarch (approved):
Looks good!

PR #83, opened 2022-06-29
Merge several nested `if` conditions into one

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims to merge several nested `if` conditions into one.

misc: use f-string instead of string concatenation x4

## Rationale

Too much nesting can make code difficult to understand, especially in Python, where there are no parentheses to help distinguish levels of nesting.

Reading deeply nested code is difficult because you have to understand which conditions refer to which levels. Therefore, we should try to avoid nesting as much as possible, and the situation where two `if` conditions can be combined using `and` is a straightforward win.

void

## Future Work

void

### DIFF
--- a/src/fprime_gds/flask/flask_uploads.py
+++ b/src/fprime_gds/flask/flask_uploads.py
@@ -191,10 +191,10 @@ def config_for_set(uset, app, defaults=None):
     if defaults is None:
         defaults = dict(dest=None, url=None)
 
-    allow_extns = tuple(config.get(prefix + "ALLOW", ()))
-    deny_extns = tuple(config.get(prefix + "DENY", ()))
-    destination = config.get(prefix + "DEST")
-    base_url = config.get(prefix + "URL")
+    allow_extns = tuple(config.get(f"{prefix}ALLOW", ()))
+    deny_extns = tuple(config.get(f"{prefix}DENY", ()))
+    destination = config.get(f"{prefix}DEST")
+    base_url = config.get(f"{prefix}URL")
 
     if destination is None and uset.default_dest:
         # use the "default_dest" callable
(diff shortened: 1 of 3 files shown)


src/fprime_gds/flask/flask_uploads.py
```diff
@@ -191,23 +191,20 @@ def config_for_set(uset, app, defaults=None):
...
-            # use the "default_dest" callable
-            destination = uset.default_dest(app)
-        if destination is None:  # still
-            # use the default dest from the config
-            if defaults["dest"] is not None:
-                using_defaults = True
-                destination = os.path.join(defaults["dest"], uset.name)
-            else:
-                 raise RuntimeError(f"no destination for set {uset.name}")
+    allow_extns = tuple(config.get(f"{prefix}ALLOW", ()))
```
LeStarch: This file is third-party code. Perhaps we should leave it as-is to maintain an upgrade path.

LeStarch (dismissed):
Reviewed, attendees: @kevin-f-ortega, @bocchino, @arizvi786, @LeStarch

LeStarch (commented):
Approved, attendees: @kevin-f-ortega, @bocchino, @arizvi786, @LeStarch

ThibFrgsGmz:
Apparently, I already modified this file two months ago during a first run on f-strings formatting.

src/fprime_gds/flask/flask_uploads.py
ThibFrgsGmz: Done. My VSCode indent-rainbow plugin spotted an indentation flaw so I took the liberty to fix it.

LeStarch (approved):

src/fprime_gds/flask/flask_uploads.py
LeStarch: That is ok!

PR #84, opened 2022-06-29
Refactor `convert()` function from `string_util.py`

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims to
- refactor the `convert` function of `string_util.py` 
- replace conditional assignment to a variable with an `if` expression
- add single value to `dictionary` rather than using `update()`.

misc: use f-string instead of string concatenation x4

## Rationale

- Swap the `if /else` statements to stop reading the function as soon as possible
- Remove the unnecessary `else` statement to limit nested statements
- list comprehension to run faster than building the collection in a loop and in one line.
--
- Python's conditional expression syntax is its version of the ternary operator. Thus, there is only one statement where `x` is defined, instead of having to read two statements plus the `if-else` lines.
--
- Eliminate the overhead of building another dictionary and calling a method.

void

## Future Work

void

### DIFF
--- a/src/fprime_gds/flask/flask_uploads.py
+++ b/src/fprime_gds/flask/flask_uploads.py
@@ -15,10 +15,7 @@ import sys
 
 PY3 = sys.version_info[0] == 3
 
-if PY3:
-    string_types = (str,)
-else:
-    string_types = (basestring,)
+string_types = (str, ) if PY3 else (basestring, )
 
 import os.path
 import posixpath
--- a/src/fprime_gds/flask/flask_uploads.py
+++ b/src/fprime_gds/flask/flask_uploads.py
@@ -15,7 +15,10 @@ import sys
 
 PY3 = sys.version_info[0] == 3
 
-string_types = (str, ) if PY3 else (basestring, )
+if PY3:
+    string_types = (str,)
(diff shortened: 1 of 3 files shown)


src/fprime_gds/flask/flask_uploads.py
```diff
@@ -15,10 +15,7 @@
 
 PY3 = sys.version_info[0] == 3
 
-if PY3:
-    string_types = (str,)
-else:
-    string_types = (basestring,)
+string_types = (str, ) if PY3 else (basestring, )
```
LeStarch: Edited third-party file.

LeStarch (commented):
Reviewed, attendees: @kevin-f-ortega, @bocchino, @arizvi786, @LeStarch.

One thing of note, we should specifically start mentioning which tests test these PR to ensure that there is CI coverage on the changes.

src/fprime_gds/flask/flask_uploads.py
ThibFrgsGmz: Modification cancelled.

LeStarch (approved):

Issue #86, opened 2022-07-13
Cannot test fresh install of F prime due to NameError: name 'Path' is not defined


rmelick-muon:
| | |
|:---|:---|
|**_F´ Version_**| fprime from `master` branch, `fprime-gds==3.1.1`|
|**_Affected Component_**|  fprime-gds  |
---
## Problem Description

I was following the F' Installation Guide, and followed the steps successfully up until "Testing F´ GDS Installation Via Running HTML GUI".

That had me run the command
```
fprime-gds -g html -r ./build-artifacts
```

which failed with the following exception
```
Traceback (most recent call last):
  File "/Users/xxx/.pyenv/versions/fprime-stock/bin/fprime-gds", line 8, in <module>
    sys.exit(main())
  File "/Users/xxx/.pyenv/versions/3.10.4/envs/fprime-stock/lib/python3.10/site-packages/fprime_gds/executables/run_deployment.py", line 277, in main
    settings = vars(get_settings())
  File "/Users/xxx/.pyenv/versions/3.10.4/envs/fprime-stock/lib/python3.10/site-packages/fprime_gds/executables/run_deployment.py", line 23, in get_settings
    root = Path(args.root_dir)
NameError: name 'Path' is not defined
```

## How to Reproduce

1. Do a fresh checkout of F', and following [installation guide](https://nasa.github.io/fprime/INSTALL.html)

## Expected Behavior

I would expect the command to work correctly, starting up the fprime-gds and allowing me to continue with the tutorial.

## Thoughts on root cause
I suspect this is caused by commit https://github.com/fprime-community/fprime-gds/commit/97f549285d6311bb2f75b6735f66da25bf4453f4, which removed the line `from pathlib import Path` from `run_deployment.py`.

## Related tickets
This was also reported in the main fprime system: https://github.com/nasa/fprime/issues/1562

rmelick-muon:
I think this commit may also have caused some other issues with a forgotten `import sys` in `executables/utils.py`

I moved forward one, but then hit this error
```
[ERROR] binary location /Users/russell/src/other/fprime/build-artifacts/Darwin/bin does not exist
Traceback (most recent call last):
  File "/Users/xxx/.pyenv/versions/fprime-stock/bin/fprime-gds", line 8, in <module>
    sys.exit(main())
[...]

PR #87, opened 2022-07-13
Fix import errors from #86

rmelick-muon:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  https://github.com/fprime-community/fprime-gds/issues/86 |

---
## Change Description

Add missing imports that caused https://github.com/fprime-community/fprime-gds/issues/86.
Add unit tests to reproduce the errors.

## Rationale

Fixes bug that prevents installation tutorial from being followed

## Future Work

Should we enable `pylama` for this project, like the main fprime project?  I think that it would have caught these errors, when I ran it locally I saw

```
src/fprime_gds/executables/run_deployment.py:23:16 E0602 undefined name 'Path' [pyflakes]
src/fprime_gds/executables/utils.py:136:9 E0602 undefined name 'sys' [pyflakes]
```

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -607,6 +607,7 @@ tcpserver
 td
 telem
 tempdir
+tempfile
 testcase
 TESTLIST
 textarea
--- a/src/fprime_gds/executables/run_deployment.py
+++ b/src/fprime_gds/executables/run_deployment.py
@@ -7,6 +7,7 @@ import os
 import sys
 import platform
 import webbrowser
+from pathlib import Path
 
 import fprime_gds.executables.cli
 import fprime_gds.executables.utils
--- a/src/fprime_gds/executables/utils.py
+++ b/src/fprime_gds/executables/utils.py
@@ -6,6 +6,7 @@ Utility functions to enable the executables package to function seamlessly.
 import atexit
 import signal
 import subprocess
+import sys
 import time
 from pathlib import Path
 from fprime.fbuild.settings import (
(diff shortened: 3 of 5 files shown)


Anirban166 (dismissed):
I like the addition of these unit tests to avoid issues like [fprime#1562](https://github.com/nasa/fprime/issues/1562) arising in future updates.

LeStarch:
Thanks for performing this fix!  We should probably have CI test by running like the install instructions too.

LeStarch:
@codeflight1, this is a great fix.  Can you make the UT run work in an empty clone?  The above failure raises a system exit during one of your new tests.

LeStarch:
Sorry, wrong submitter.  @rmelick-muon, can you look into the failing test.

rmelick-muon:
@LeStarch good call, I hadn't seen the test failure, but was easily able to reproduce locally and fix it.  Any advice on the `tempfile` spelling failure (it's a python module)?  Should I run the command specified in the comment on this PR, or manually add it to `.github/actions/spelling/expect.txt`?

Also, I really like your idea of actually running the install commands from a shell, not just as a python unit test.  Those commands would require main `fprime` to be checked out and built.  I could see wanting to do that inside the main `fprime` repository as well as here.  That way `fprime-gds` would know if its changes were about to break `fprime`, and `fprime` would know if their dependency updates to a newer `fprime-gds` would break.

Do you think we should add the step to check out and build `fprime` to `fprime-gds`, or add a step to run `fprime-gds` to the `fprime` github workflows, or both?

LeStarch:
@rmelick-muon add the spelling fix to `expect.txt`.  At some point we need to do better configuration handling with the spelling action, but for now it keeps us from generating spelling issues over time.

LeStarch:
@rmelick-muon let's get this PR in and then look into a better CI action.  I found another bug last night in the GDS that is very similar to the one you are fixing here....so I'd like to get this in soon.

LeStarch (approved):
Looks good!

rmelick-muon:
sounds great!  thanks for the help :)

PR #92, opened 2022-08-20
Refactor `seq_file_parser.py` for performance and readability

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims to:
- Remove the unnecessary else statement to limit nested statements (`def parseAbsolute` L210)
- Convert a `for` loop into a `list` comprehension and directlt returning it (`def replaceSpacesAndCommas` L69)
- Use f-string instead of string interpolation
- Replace `dict()` / `list()` with `{}` / `[]` for performance speed

## Rationale

Clean code, better performance and readability.

void

## Future Work

void

### DIFF
--- a/src/fprime_gds/common/parsers/seq_file_parser.py
+++ b/src/fprime_gds/common/parsers/seq_file_parser.py
@@ -201,7 +201,8 @@ class SeqFileParser:
                             descriptor, seconds, useconds = parseTime(i, line[0])
                         except Exception:
                             raise gseExceptions.GseControllerParsingException(
-                                f"Line {i + 1}: Encountered syntax error parsing timestamp")
+                                f"Line {i + 1}: Encountered syntax error parsing timestamp"
+                            )
                         mnemonic = line[1]
                         args = []
                         if length > 2:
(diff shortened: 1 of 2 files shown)


src/fprime_gds/common/parsers/seq_file_parser.py
```diff
@@ -85,25 +82,24 @@ def parseArg(arg):
...
-                            # See if it translates to a float:
-                            return float(arg)
-                        except ValueError:
-                            # Otherwise it is an enum type:
-                            return str(arg)
+                        # Otherwise it is an enum type:
+                        return str(arg)
 
-            return list(map(parseArg, args))
+            return [map(parseArg, args)]
```
LeStarch: I think this line is wrong.  This creates a list of one item which is the object returned by the `map` call.

list(map(...)) instead iterates the result of the map call and makes it a `list` type.  It is equivalent to:

```
[item for item in map(...)]
```

ThibFrgsGmz: Done!

LeStarch (approved):
Fixes review notes.  I am curious why you use a list comprehension as opposed to the original `list(map())` call as it is longer, but the inheritance issue has been fixed.

ThibFrgsGmz:
> @LeStarch I am curious why you use a list comprehension as opposed to the original list(map()) call as it is longer

Since I learned that it is slower to use `dict`/`list` keywords, I tend to always want to delete them.

LeStarch (approved):

PR #93, opened 2022-08-20
Refactor `telemetry_predicate.__call__` to save mental load

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims to
- refactor the `telemetry_predicate.__call__` function of `predicated.py` with boolean `if` expression identity.

## Rationale

- Instead of using an if expression to evaluate a boolean, you can simply use it directly. This is shorter and clearer. Indeed if-expressions of this form take some time to be mentally analyzed.

void

## Future Work

void

### DIFF
--- a/src/fprime_gds/common/testing_fw/predicates.py
+++ b/src/fprime_gds/common/testing_fw/predicates.py
@@ -535,15 +535,17 @@ class telemetry_predicate(predicate):
         Args:
             telemetry: an instance of ChData (object)
         """
-        if not isinstance(telemetry, ChData):
-            return False
-        if (
-            self.id_pred(telemetry.get_id())
-            and self.value_pred(telemetry.get_val())
-            and self.time_pred(telemetry.get_time())
-        ):
-            return True
-        return False
+        return (
+            bool(
+                (
+                    self.id_pred(telemetry.get_id())
+                    and self.value_pred(telemetry.get_val())
+                    and self.time_pred(telemetry.get_time())
+                )
+            )
+            if isinstance(telemetry, ChData)
+            else False
+        )
 
     def __str__(self):
         """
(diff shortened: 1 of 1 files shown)


src/fprime_gds/common/testing_fw/predicates.py
```diff
@@ -535,15 +535,17 @@ def __call__(self, telemetry):
...
-            return False
-        if (
-            self.id_pred(telemetry.get_id())
-            and self.value_pred(telemetry.get_val())
-            and self.time_pred(telemetry.get_time())
-        ):
-            return True
-        return False
+        return (
+            bool(
```
LeStarch: Is the `bool` here necessary? It seems the results of the `and` block would always be a bool.  Also shouldn't we just lead with `if  isinstance(telemetry, ChData)`?  e.g.

```
return isinstance(telemetry, ChData) and self.id_pred(telemetry.get_id()) and ... ...
```
Short-circuit evaluation should prevent invalid execution...

LeStarch (changes requested):

src/fprime_gds/common/testing_fw/predicates.py
ThibFrgsGmz: Done!

LeStarch:
Looks like the tests are failing because of a missed `)` at the end.  Otherwise I like the change!

ThibFrgsGmz:
@LeStarch It seems to me that I have made the requested change, right? Is there any action I should take?

src/fprime_gds/common/testing_fw/predicates.py
```diff
@@ -1,4 +1,4 @@
-"""
+"""f
```
LeStarch: Why did this 'f' appear?  Otherwise, I am happy.

ThibFrgsGmz: Hmmm I don't know... How I didn't see it. 🤔 I just removed it!!

LeStarch (approved):

LeStarch (approved):

PR #101, opened 2022-11-03
Replace LGTM with Github Actions

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**| GitHub Actions |
|**_Related Issue(s)_**|  |

---
## Change Description

Adds a Github Actions workflow that runs the security and quality static code analysis in order to phase LGTM out.

## Rationale

[lgtm.com](https://github.com/nasa/fprime/pull/lgtm.com) will be shut down at the end of the year.

## Future Work

Remove the LGTM configuration files once we're ok with phasing it out.

## Comments

Results should be viewable once the checks are done. It's currently running all queries from the security-and-quality suite, let me know if we want to filter some categories out.

### DIFF
--- a/.github/actions/codeql/security-pack.yml
+++ b/.github/actions/codeql/security-pack.yml
@@ -1,0 +1,4 @@
+name: "CodeQL security and quality"
+
+queries: 
+  - uses: security-and-quality
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -413,6 +413,7 @@ ninjaaron
 nitpicky
 noapp
 noqa
+normpath
 nosort
 Noto
 novalidate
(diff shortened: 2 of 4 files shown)


LeStarch (dismissed):
The changes look good, however; we might fix the breakages. @thomas-bc do you want to do that in this PR, or a separate one?

thomas-bc:
I implemented the recommended way of dealing with the `Uncontrolled data used in path expression` [alerts](https://github.com/fprime-community/fprime-gds/security/code-scanning/2). Let me know what you think. It's using [normpath()](https://docs.python.org/3/library/os.path.html#os.path.normpath) to normalize the path (i.e. resolve special instructions like `../` etc...) and check that the requested resource is within logdir.

The two `Wrong number of arguments in a call` [alerts](https://github.com/fprime-community/fprime-gds/security/code-scanning/15) seem to be expected in both cases, so we should probably dismiss those alerts. One has the LGTM suppression flag, the other one is a test case testing the raising of an exception.

LeStarch:
@thomas-bc looks good to me.  Can we suppress-by-comment or is there some other way to do it?  Regardless, I agree with the proposed path-forward.

LeStarch (approved):
@thomas-bc approved!

PR #103, opened 2022-11-16
Pass repeated condition to own statement

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| None  |
|**_Affected Architectures(s)_**| None  |
|**_Related Issue(s)_**| None |

---
## Change Description

This PR aims to pass repeated conditions to their own `if` instruction.

## Rationale

Duplicating conditions in an `if/elseif` statement makes things more difficult to read and causes code duplication issues.

Although this increases the number of nested `if` statements, it improves the understanding of the code at first glance.

P.S. I took the opportunity to run `isort` on the two files I modified

None

## Future Work

None

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -18,11 +18,10 @@ import re
 import sys
 
 import fprime_gds.common.communication.adapters.base
-import fprime_gds.common.communication.checksum
-import fprime_gds.common.logger
-
 # Include basic adapters
 import fprime_gds.common.communication.adapters.ip
+import fprime_gds.common.communication.checksum
+import fprime_gds.common.logger
 import fprime_gds.common.utils.config_manager
 
 try:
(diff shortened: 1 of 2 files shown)


src/fprime_gds/executables/cli.py
```diff
@@ -501,10 +500,11 @@ def handle_arguments(cls, args, **kwargs):
 
         # Handle configuration arguments
         config = fprime_gds.common.utils.config_manager.ConfigManager()
-        if args.config is not None and os.path.isfile(args.config):
-            config.set_configs(args.config)
-        elif args.config is not None:
-            raise ValueError(f"Configuration file {args.config} not found")
+        if args.config is not None:
```
LeStarch, line 503: In the python in mainline, we had issue with the if-depth of this kind of refactor.  This seems to be in the same category of changes.  Thoughts?

LeStarch (approved):

LeStarch:
In general this seems cleaner to read to me, but we had different thoughts on a core F´ PR.  Just wondering on that before I merge.

LeStarch (approved):

src/fprime_gds/executables/cli.py
ThibFrgsGmz: I don't like nested statements either, but it seemed to me that these were more astute (code duplication and readability).

PR #107, opened 2022-11-18
Refactor unneeded `else` / `elif` when the `if` block has a `break`/`continue`/`return`/`raise` statement

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**| None |
|**_Affected Architectures(s)_**| None |
|**_Related Issue(s)_**| None |

---
## Change Description

This PR aims to refactor unneeded `else` / `elif` when the `if` block has a `break`/`continue`/`return`/`raise` statement.

## Rationale

The use of `else` or `elif` becomes redundant and can be dropped `if` the last statement under the main `if` / `elif` block is a `break`/`continue`/`raise`/`return` statement.

In the case of an `elif` after these statements, it can be written as a separate if block.
For `else` blocks after these statements, the statements can be moved out of the `else`. 

Shuffling the code in this way improves its readability and makes it easier to maintain.

None

## Future Work

None

### DIFF
--- a/src/fprime_gds/common/communication/framing.py
+++ b/src/fprime_gds/common/communication/framing.py
@@ -174,11 +174,10 @@ class FpFramerDeframer(FramerDeframer):
                 ):
                     data = data[total_size:]
                     return deframed, data
-                else:
-                    print(
-                        "[WARNING] Checksum validation failed. Have you correctly set '--comm-checksum-type'",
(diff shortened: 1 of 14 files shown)


LeStarch (approved):

LeStarch:
Reviewed with @timcanham and it looks good.

PR #111, opened 2023-01-05
Update/fix type problem

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes command arguments for types that begin with 'F' and aren't F32 nor F64.

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -456,7 +456,6 @@ class DictionaryParser(DetectionParser):
 
     def handle_arguments(self, args, **kwargs):
         """ Handle arguments as parsed """
-        args = super().handle_arguments(args, **kwargs)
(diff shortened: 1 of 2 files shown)


src/fprime_gds/executables/cli.py
```diff
@@ -456,7 +456,6 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
 
     def handle_arguments(self, args, **kwargs):
         """ Handle arguments as parsed """
-        args = super().handle_arguments(args, **kwargs)
```
LMPS97, line 459: Is it a concern that `super().handle_arguments(args, **kwargs)` is not being called anymore on this function?

LMPS97 (approved):

PR #125, opened 2023-04-22
Add some refactoring

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

- Replace if statement with if expression
- Remove string casting
- Some f-string formatting
- Swap one if expression to remove a negation
- Move a guard from `while` statement for its test conditions

## Rationale

- Use better patterns
- Reduce technical debt by improving code quality

Let's see CI output 

P.S. I had laziness to make a PR by type of refactoring - sorry.

## Future Work

None

### DIFF
--- a/src/fprime_gds/common/data_types/ch_data.py
+++ b/src/fprime_gds/common/data_types/ch_data.py
@@ -13,6 +13,7 @@ from fprime_gds.common.utils.string_util import format_string_template
 from fprime.common.models.serialize.serializable_type import SerializableType
 from fprime.common.models.serialize.array_type import ArrayType
 
+
 class ChData(sys_data.SysData):
     """
     The ChData class stores a specific channel telemetry reading.
@@ -63,7 +64,11 @@ class ChData(sys_data.SysData):
         # in which case we just display the description
         if val_obj is None:
             return template.ch_desc
-        temp_val = val_obj.val if not isinstance(val_obj, (SerializableType, ArrayType)) else val_obj.formatted_val
+        temp_val = (
+            val_obj.val
+            if not isinstance(val_obj, (SerializableType, ArrayType))
+            else val_obj.formatted_val
+        )
         fmt_str = template.get_format_str()
         if temp_val is None:
             return ""
--- a/src/fprime_gds/common/data_types/ch_data.py
+++ b/src/fprime_gds/common/data_types/ch_data.py
@@ -67,7 +67,9 @@ class ChData(sys_data.SysData):
         fmt_str = template.get_format_str()
         if temp_val is None:
             return ""
-        return format_string_template(fmt_str, (temp_val,)) if fmt_str else temp_val
+        if fmt_str:
+            return format_string_template(fmt_str, (temp_val,))
+        return temp_val
 
     def set_pkt(self, pkt):
         """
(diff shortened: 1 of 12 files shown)


src/fprime_gds/common/data_types/ch_data.py
```diff
@@ -67,9 +67,7 @@ def _compute_display_text(self, val_obj, template):
         fmt_str = template.get_format_str()
         if temp_val is None:
             return ""
-        if fmt_str:
-            return format_string_template(fmt_str, (temp_val,))
-        return temp_val
+        return format_string_template(fmt_str, (temp_val,)) if fmt_str else temp_val
```
thomas-bc: I would argue that this is less readable than before the change, and therefore not a better pattern. A succession of if statements clearly shows the priority order. Here we would have one if statement, then one if expression -- the reader kind of has to unfold that in their head and it takes a couple seconds to do so.

src/fprime_gds/common/decoders/file_decoder.py
```diff
@@ -57,8 +57,4 @@ def decode_api(self, data):
         if packetType == "END":  # Packet Type is END
             hashValue = struct.unpack_from(">I", data, 5)
             return [file_data.EndPacketData(seqID, hashValue),]
-        if packetType == "CANCEL":  # Packet Type is CANCEL
-            # CANCEL Packets have no data
-            return [file_data.CancelPacketData(seqID),]
-        # The data was not determined to be any of the packet types so return none
-        return None
+        return [file_data.CancelPacketData(seqID),] if packetType == "CANCEL" else None
```
thomas-bc: Same as above. If statements are nice but I (personally) don't think they should be used everywhere they can be

thomas-bc (changes requested):
I added a couple comments, let me know what you think and if you're ok reverting the changes. Other stuff looks good to me, thanks for improving our code quality!

src/fprime_gds/common/data_types/ch_data.py
ThibFrgsGmz: As you want, there already is an if expr just above. But I respect your choice - you are the maintainer 😸

By the way, both files where you put a comment arenot black formatted, is it intended?

src/fprime_gds/common/decoders/file_decoder.py
ThibFrgsGmz: Done.

src/fprime_gds/common/data_types/ch_data.py
thomas-bc: Ah yes, I think there's a reason we're not running formatting checks on this repo, but I will fix that. Feel free to format those files you're modifying!
We want to avoid formatting the whole codebase at once to avoid large commits/PRs. So we're progressively formatting as things get modified.

ThibFrgsGmz: Done chef!

thomas-bc: Thank you very much 😄

LeStarch:
@thomas-bc want to re-review? I see the requests were outdated/resolved.

thomas-bc (approved):
This is good stuff, thanks!

PR #127, opened 2023-04-29
Refactor containers iteration for performance

v1k1nghawk:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

- Refactor iterating over containers
- Remove trailing whitespaces

## Rationale

- Performance optimization
- Clean code

## Future Work

None

### DIFF
--- a/src/fprime_gds/common/history/ram.py
+++ b/src/fprime_gds/common/history/ram.py
@@ -90,7 +90,7 @@ class RamHistory(History):
             if len(self.retrieved_cursors.values()) > 0:
                 earliest = min(self.retrieved_cursors.values())
             del self.objects[:earliest]
-            for key in self.retrieved_cursors.keys():
+            for key in self.retrieved_cursors:
                 self.retrieved_cursors[key] -= earliest
 
     def sessions(self):
--- a/src/fprime_gds/common/parsers/seq_file_parser.py
+++ b/src/fprime_gds/common/parsers/seq_file_parser.py
@@ -99,7 +99,7 @@ class SeqFileParser:
                         # Otherwise it is an enum type:
                         return str(arg)
 
-            return [item for item in map(parseArg, args)]
+            return list(map(parseArg, args))
 
         def parseTime(lineNumber, time):
             """
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -295,7 +295,7 @@ class CommAdapterParser(ParserBase):
                 "type": str,
                 "help": "Adapter for communicating to flight deployment. [default: %(default)s]",
                 "choices": ["none"]
-                + [name for name in adapter_definition_dictionaries.keys()],
+                + list(adapter_definition_dictionaries),
                 "default": "ip",
             },
             ("--comm-checksum-type",): {
--- a/src/fprime_gds/common/decoders/event_decoder.py
+++ b/src/fprime_gds/common/decoders/event_decoder.py
@@ -59,7 +59,7 @@ class EventDecoder(decoder.Decoder):
             or None if the data is not decodable
         """
         ptr = 0
-        
+
         event_list = []
 
         while (ptr < len(data)):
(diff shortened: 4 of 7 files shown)


LeStarch:
@ThibFrgsGmz any chance you would weigh in here?  In the past you have submitted similar PRs and would like your insight into these changes.

LeStarch (commented):
Functionally, the code submitted here should operate identically to the original!

I don't know the relative performance of the list comprehension vs the `list()` function call.

thomas-bc (approved):
LGTM!
For the `list()` vs comprehension matter, `list()` seems to be more optimal indeed (see [here](https://stackoverflow.com/a/29356931)). However we're only iterating over a couple of `args` at most so performance should not be a decision factor here in my opinion.
That being said, I agree with the change!

ThibFrgsGmz:
Can only do it tonight (in 10h - NZ time) as it's not convenient on Github Mobile App.

LeStarch:
@ThibFrgsGmz if you are willing to take a look, then I am willing to wait until you have it done as you provide good insight.

src/fprime_gds/common/history/ram.py
```diff
@@ -90,7 +90,7 @@ def clear(self, start=None):
             if len(self.retrieved_cursors.values()) > 0:
                 earliest = min(self.retrieved_cursors.values())
             del self.objects[:earliest]
-            for key in self.retrieved_cursors.keys():
+            for key in self.retrieved_cursors:
```
ThibFrgsGmz, line 93: Yes, since the default behaviour when iterating over a dictionary is to iterate over the keys, the call to `keys()` is not necessary.

src/fprime_gds/common/parsers/seq_file_parser.py
```diff
@@ -99,7 +99,7 @@ def parseArg(arg):
                         # Otherwise it is an enum type:
                         return str(arg)
 
-            return [item for item in map(parseArg, args)]
+            return list(map(parseArg, args))
```
ThibFrgsGmz, line 102: Yes, much more concise and in a functional programming sense. The new version is more concise and a bit more efficient by converting the iterable returned by map directly into a list.

src/fprime_gds/executables/cli.py
```diff
@@ -295,7 +295,7 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
                 "type": str,
                 "help": "Adapter for communicating to flight deployment. [default: %(default)s]",
                 "choices": ["none"]
-                + [name for name in adapter_definition_dictionaries.keys()],
+                + list(adapter_definition_dictionaries),
```
ThibFrgsGmz, line 298: Nice one!

ThibFrgsGmz (approved):
LGTM!

Review conducted as scheduled 10 hours later ahah :D

src/fprime_gds/common/parsers/seq_file_parser.py
ThibFrgsGmz: As for the effectiveness, I'm not so sure. I tried this:

```sh
python3 -m timeit --number=3000000 "
def square(number):
    return number ** 3
numbers = [1, 2, 3, 4, 5]
list(map(square, numbers))
"
3000000 loops, best of 5: 1.68 usec per loop
```

and 

```sh
python3 -m timeit --number=3000000 "
numbers = [1, 2, 3, 4, 5]
[num ** 3 for num in numbers]
"
3000000 loops, best of 5: 1.49 usec per loop
```

But this may be a bias, as the benchmark also includes the definition of the function. Any thoughts?

PS: I still prefer the functional programming style because it is clearer, I just want to understand the finding.

v1k1nghawk: @ThibFrgsGmz Thank you for the benchmarking.

I suppose that a fairer comparison in this case would be between this:

```python
python3 -m timeit --number=3000000 "
def square(number):
       return number ** 3
numbers = [1, 2, 3, 4, 5]
list(map(square, numbers))
"
3000000 loops, best of 5: 1.95 usec per loop
```

and

```python
python3 -m timeit --number=3000000 "
def square(number):
    return number ** 3
numbers = [1, 2, 3, 4, 5]
[item for item in map(square, numbers)]
"
3000000 loops, best of 5: 2.14 usec per loop
```

Of course, the performance diff isn't huge, but the proposed code is a bit faster.

LeStarch:
@ThibFrgsGmz @thomas-bc thanks for the analysis!  I will merge this given your approvals.

@v1k1nghawk thanks for the submission!

PR #128, opened 2023-05-05
Downlink files are placed in /tmp/<username>/fprime-downlink

gjwatney:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Downlinked files are placed in /tmp/<username>/fprime-downlink

## Rationale

This fixes the problem described in https://github.com/nasa/fprime/issues/1970

Let different users try and download files.  The GDS should not throw a Permission Error

## Future Work

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -250,6 +250,7 @@ getmembers
 getpid
 getroot
 getsize
+getuser
 github
 globals
 gmail
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -15,6 +15,7 @@ import os
 import re
 import platform
 import sys
+import getpass
 
 import fprime_gds.common.logger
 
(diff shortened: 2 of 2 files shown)


thomas-bc:
@gjwatney I'm not sure I understand the reason you're choosing /tmp/_username_ here. Are you expecting the permissions to be different there than in /tmp ? I'm thinking this might be specific to your system. 
As was mentioned in #1970 a good approach could be to handle the case where a permission error occurs and keep the process going instead of crashing.

LeStarch:
The primary issue is if multiple users run on one shared GDS box, then their instances of the GDS will collide.

LeStarch:
@thomas-bc for now I like this fix.  We at some point should consider the following:

1. A better place to log: temp directory from "tempfile" package.  This will work, but is hard to find.  "logs/downlink" to keep it with the other gds logs.
2.  Fix the permissions error in-case.  However, fixing it here would not help the root issue as the file would not downlink.

LeStarch (approved):

thomas-bc (approved):

PR #129, opened 2023-05-12
Add `flake8-error-message` convention

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This feature implements the [flake8-err-msg](https://pypi.org/project/flake8-errmsg/) which allows the formatting of nice error messages.

## Rationale

The problem is that Python includes the line with the lift in the default traceback. This means that a user receives a message like this:

```py
sub = "Some value"
raise RuntimeError(f"{sub!r} is incorrect")
```

```sh
Traceback (most recent call last):
  File "tmp.py", line 2, in <module>
    raise RuntimeError(f"{sub!r} is incorrect")
RuntimeError: 'Some value' is incorrect
```

If it is longer or more complex, the duplication can be confusing for a user not used to reading tracebacks.

On the other hand, if you always assign to something like msg, you get :

```py
sub = "Some value"
msg = f"{sub!r} is incorrect"
raise RunetimeError(msg)
```

```sh
Traceback (most recent call last):
  File "tmp.py", line 3, in <module>
    raise RuntimeError(msg)
RuntimeError: 'Some value' is incorrect
```

There is now a more straightforward trace, less code and no double messages.

[Ref - Ruff linter](https://beta.ruff.rs/docs/rules/f-string-in-exception/)

Inspection is enough

## Future Work

None

### DIFF
--- a/src/fprime_gds/common/communication/adapters/ip.py
+++ b/src/fprime_gds/common/communication/adapters/ip.py
@@ -34,7 +34,8 @@ def check_port(address, port):
         socket_trial = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         socket_trial.bind((address, port))
     except OSError as err:
-        raise OSError(f"Error with address/port of '{address}:{port}' : {err}")
+        msg = f"Error with address/port of '{address}:{port}' : {err}"
+        raise OSError(msg)
     finally:
         if socket_trial is not None:
             socket_trial.close()
--- a/src/fprime_gds/common/communication/adapters/uart.py
+++ b/src/fprime_gds/common/communication/adapters/uart.py
@@ -172,17 +172,20 @@ class SerialAdapter(fprime_gds.common.communication.adapters.base.BaseAdapter):
         """
         ports = map(lambda info: info.device, list_ports.comports(include_links=True))
         if args["device"] not in ports:
+            msg = f"Serial port '{args['device']}' not valid. Available ports: {ports}"
             raise ValueError(
-                f"Serial port '{args['device']}' not valid. Available ports: {ports}"
+                msg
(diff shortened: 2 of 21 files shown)


thomas-bc (commented):
Changes LGTM.

If this is a recommended standard I won't object to it at all, but just for the sake of raising a debate (which is something I like to do, in case you haven't realized by now 😄):
98% of the cases I agree that it declutters and this is a good thing, but there are edge cases where seeing the f-string can actually be a good thing, no?

_e.g._ say your args are badly formatted or something, and `device` is empty. With this change, you would get an error like this
```
raise ValueError(
                msg
            )
ValueError: Serial port ' ' not valid. Available ports:
```
Which really doesn't say much. Whereas if you keep the f-string in the raise statement, you get 
```
raise ValueError(
                f"Serial port '{args['device']}' not valid. Available ports: {ports}"
)
ValueError: Serial port ' ' not valid. Available ports:
```
and this help to immediately identify where the issue is coming from. Thoughts? Do you know if flake8 addresses this?

thomas-bc (approved):

ThibFrgsGmz:
Sorry for the delay. My computer has been down since Sunday morning (just after my PR as it happens...) and will probably be fixed this weekend when I can look more carefully at your comment!

LeStarch (approved):

LeStarch:
We agreed that these changes are good, however, as a note: in F´  we expect to trap exceptions and present a clean error message to users without a confusing stack-trace. Thus the motivation here becomes one of appeasing a potentially good linter and less to beautify messages for users directly.

PR #131, opened 2023-05-15
Fixing ids argument

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

The `--ids` parameter was converted to an int but an iterable was requested for the filter. I changed the type of the param to fix the error and support multiple ids.

The `nargs` is documented [here](https://docs.python.org/3/library/argparse.html?highlight=argparse#nargs).

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -725,6 +725,7 @@ class SearchArgumentsParser(ParserBase):
                 "action": "store",
                 "required": False,
                 "type": int,
+                "nargs":'+',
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):
Argh, that's my bad, seems I missed the nargs when I moved that around in 6846ece7c812d985a387fa5b0cec74ca3f31d061. (for reference, see [here](https://github.com/fprime-community/fprime-gds/commit/6846ece7c812d985a387fa5b0cec74ca3f31d061#diff-9924ca0764e6bd9dbf5d1320b29d53f8201cbd6e451912602727195dc9d96554L88)).
Thanks for catching that and making the fix!
LGTM.

PR #134, opened 2023-07-14
Adding safer JSON parsing for NaN Infinity and null

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes a bug where Python's JSON supports NaN and Infinity values, but JavaScript does not.  This fix replaces those references with a string representation and then replaces them back in post.

### DIFF
--- a/src/fprime_gds/flask/static/js/loader.js
+++ b/src/fprime_gds/flask/static/js/loader.js
@@ -67,7 +67,7 @@ class SaferParser {
             UNQUOTED: 0,
             QUOTED: 1
         };
-        this.FLAG = "-_-";
+        this.FLAG = "-_-您好"; // Extended character usage make collisions less-likely
         this.MAPPINGS = [
             ["-Infinity", this.FLAG + "-inf", -Infinity],
             ["Infinity", this.FLAG + "inf", Infinity],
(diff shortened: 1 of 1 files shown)


src/fprime_gds/flask/static/js/loader.js
```diff
@@ -49,6 +50,118 @@ export function loadTextFileInputData(event) {
...
+            UNQUOTED: 0,
+            QUOTED: 1
+        };
+        this.FLAG = "-_-您好"; // Extended character usage make collisions less-likely
+        this.MAPPINGS = [
+            ["-Infinity", this.FLAG + "-inf", -Infinity],
+            ["Infinity", this.FLAG + "inf", Infinity],
+            ["NaN", this.FLAG + "nan", NaN],
+            ["null", this.FLAG + "null", null]
+        ];
```
jwest115, line 76: @LeStarch why are we using a flag here? Is it so that if "-inf", "inf", "nan", or "null" appear in a string we could differentiate between that and the substring converted by the SafeParser?

LeStarch: Yes.  A better flag-free solution would have required changes on both the Python end and here.  It seemed a localized solution converting to something safe and immediately back was a better solution because:

1. Colocated changes
2. Does not increase size of transmitted data
3. These cases should be sufficiently rare as to not merit a much more complex and comprehensive solution

jwest115: thanks for the explanation! looks good to me!

jwest115 (approved):

PR #135, opened 2023-08-03
Add display text if there is no format string

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| data_types  |
|**_Related Issue(s)_**|  |

---
## Change Description

This merge request add a `display_text` for events having arguments but not a `format_string`. 

## Rationale

Before the arguments of an event without a `format_string` were not logged or printed everywhere. This pull request will enable the visualization of the argument also if the `format_string` is not set.

An event with no `format_string` but three arguments(an integer, an enum and a float) will contain in the column Event Description of gds: `[{'My_integer_arg': 123}, {'My_enum_arg': 'VALUE_OF_ENUM'}, {'My_float_arg': 12.34}]`.

### DIFF
--- a/src/fprime_gds/common/data_types/event_data.py
+++ b/src/fprime_gds/common/data_types/event_data.py
@@ -43,7 +43,8 @@ class EventData(sys_data.SysData):
         if event_args is None:
             self.display_text = event_temp.description
         elif event_temp.format_str == "":
-            self.display_text = str([arg.val for arg in event_args])
+            args_template = self.template.get_args()
+            self.display_text = str([{args_template[index][0]:arg.val} for index, arg in enumerate(event_args)])
         else:
             self.display_text = format_string_template(
                 event_temp.format_str, tuple([arg.val for arg in event_args])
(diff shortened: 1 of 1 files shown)


bocchino:
Can you explain the motivation for this PR a bit more?

1. Event specifiers in FPP are required to have format strings. What is the scenario where an event specifier without a format string is presented to the GDS?
2. What is the reason for handling just this case of an ill-specified event (n > 0 args and no format specifier)? What about an event that has n args and m replacement fields in its format specifier, for n =/= m? This is also disallowed by FPP.

SMorettini:
About point 1: I thought about this change because I was using gds to communicate with a subsystem of my project that doesn't use Fprime. So I created the xml dictionary manually and I didn't put the format string. 
About point 2: I didn't consider that case. 

If the idea is to support only the dictionary produced by fpp, then we can close this pull request directly.

LeStarch:
I see this PR as patching an edge case.  What to do when format-string == ""?  In FPP the user would have to explicitly place an empty string which likey means that was intended...but that is not useful.

LeStarch:
Outside of FPP (dictionary) it means the text was hand-edited and the format string was forgotten.

LeStarch (approved):
Rationale: if a user explicitly wants an empty format string then " " may be used.  This fixes the stated issue and prints something if the format string was forgotten.

thomas-bc (approved):

PR #136, opened 2023-08-08
Initial setup of fprime-gds to work with OpenMCT

mohitsingh999:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Mohit Singh |
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**| F-Prime GDS  |
|**_Related Issue(s)_**|  |

---
## Change Description

Add functionality to run OpenMCT from the F-Prime GDS. A user can specify the openmct argument at command line by running `fprime-gds --openmct` , and they will get two tabs; one that launches the standard GDS, and one that launches an OpenMCT server with all the telemetry channels registered and showing Real Time Telemetry. 

## Rationale
The current F-Prime GDS has two ways of visualizing telemetry. `Channels` and `Charts`. These telemetry visualizations can be expanded upon by integrating OpenMCT with F-Prime. OpenMCT allows for users to create dashboards, utilize gauges, stacked/overlay plots, and tables. It also incorporates conditional coloring based on user defined limits.  

Currently this functionality has been fully tested with the base `Ref` deployment. It should be tested with additional deployments to ensure no critical values are hardcoded, since that would hinder general usability.

**NOTE**: This pull request is breaking. Users will need to install the `fprime_openmct` python package repo since that is imported at the start of `run_deployment.py`. That package is available at https://github.com/mohitsingh999/fprime-openmct currently and can be registered as a python package by running `pip install .` at the root of `fprime-openmct`. A pull request has been created to merge this repo into https://github.com/fprime-community/fprime-openmct, and is currently being reviewed. 

## Future Work

Work needs to be done to ensure the OpenMCT integration with `fprime-gds` happens in parallel with integration of `fprime-openmct` in `fprime-community`. General installation instructions will need to be updated to inform users to install `fprime_openmct` with pip, in addition to `fprime_tools` and `fprime_gds`. `fprime-openmct` needs to be registered on PyPI as well, just like `fprime-gds` and `fprime-tools`. 

@jwest115 will be taking over managing OpenMCT integration related tasks after August 10th

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -782,3 +782,41 @@ class RetrievalArgumentsParser(ParserBase):
 
     def handle_arguments(self, args, **kwargs):
         return args
+    
+class OpenMCTTelemetryPollerParser(ParserBase):
+    """ Parser for OpenMCT Server arguments  """
+
+    def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
+        """ Return arguments on whether to use OpenMCT, post to a specific URI, and publish telemetry at a specific frequency"""
+        return {
+            ("--openmct",): {
+                "dest": "openmct",
+                "action": 'store_true',
+                "help": "Determines whethers OpenMCT will be used for Telemetry Visualization",
+                "required": False
+            },
+            ("--openmct-uri",): {
+                "dest": "openmct_uri",
+                "type": str,
+                "default": "http://127.0.0.1:4052/fprime_telem",
+                "help": "URI of the OpenMCT Server. The URI at which the F-Prime telemetry will be broadcasted.",
+                "required": False
+            },
+            ("--openmct-telem-rate",): {
+                "dest": "openmct_telem_rate",
+                "type": float,
+                "default": 1,
+                "help": "Rate(in Hz) at which we want to poll the F-Prime Telemetry Pipeline for new telemetry",
+                "required": False
+            },
+            ("--openmct-dir",): {
+                "dest": "openmct_dir",
+                "type": str, 
+                "default": "",
+                "help": "Directory where the OpenMCT Installation is located",
+                "required": False
+            }
+        }
+
+    def handle_arguments(self, args, **kwargs):
+        return args
--- a/src/fprime_gds/executables/run_deployment.py
+++ b/src/fprime_gds/executables/run_deployment.py
@@ -19,10 +19,14 @@ from fprime_gds.executables.utils import AppWrapperException, run_wrapped_applic
 
 # Try to Import FPrime-OpenMCT Python Packages
 try: 
+    import fprime_openmct
+except ImportError:
+    fprime_openmct = None
+
+if fprime_openmct is not None: 
     from fprime_openmct.config_server import ServerConfig
     from fprime_openmct.fprime_to_openmct import TopologyAppDictionaryJSONifier
-except:
-    pass
+
 
 
 BASE_MODULE_ARGUMENTS = [sys.executable, "-u", "-m"]
@@ -211,6 +215,9 @@ def main():
 
     # Check if OpenMCT is set to be used. If true, generate OpenMCT States and Launch the OpenMCT Server
     if parsed_args.openmct:
+
+        if fprime_openmct is None:
+            raise ImportError('FPrime-OpenMCT Bridge not installed. Please install fprime_openmct with pip.')
 
         # Try Generating the OpenMCT JSON and Initial States
         top_dict = TopologyAppDictionaryJSONifier(str(parsed_args.dictionary))
--- a/src/fprime_gds/executables/run_deployment.py
+++ b/src/fprime_gds/executables/run_deployment.py
@@ -19,7 +19,6 @@ from fprime_gds.executables.utils import AppWrapperException, run_wrapped_applic
 
 # Try to Import FPrime-OpenMCT Python Packages
 try: 
-    import fprime_openmct 
     from fprime_openmct.config_server import ServerConfig
     from fprime_openmct.fprime_to_openmct import TopologyAppDictionaryJSONifier
 except:
@@ -166,10 +165,9 @@ def launch_openmct(parsed_args):
     Return:
         launched process
     """
-    openmct_dir = fprime_openmct.__file__.replace('__init__.py', 'javascript')
     openmct_server = ServerConfig()
 
-    app_cmd = openmct_server.launch_openmct_server(openmct_dir)
+    app_cmd = openmct_server.launch_openmct_server()
 
     return app_cmd 
 
@@ -215,10 +213,9 @@ def main():
     if parsed_args.openmct:
 
         # Try Generating the OpenMCT JSON and Initial States
-        openmct_dir = fprime_openmct.__file__.replace('__init__.py', 'javascript')
         top_dict = TopologyAppDictionaryJSONifier(str(parsed_args.dictionary))
-        top_dict.writeOpenMCTJSON('FPrimeDeploymentTopologyAppDictionary', openmct_dir)
-        top_dict.writeInitialStatesJSON('initial_states', openmct_dir)
+        top_dict.writeOpenMCTJSON('FPrimeDeploymentTopologyAppDictionary')
+        top_dict.writeInitialStatesJSON('initial_states')
 
         # Try Launching OpenMCT Server
         launchers.append(launch_openmct)
--- a/src/fprime_gds/executables/run_deployment.py
+++ b/src/fprime_gds/executables/run_deployment.py
@@ -19,7 +19,7 @@ from fprime_gds.executables.utils import AppWrapperException, run_wrapped_applic
 
 import fprime_openmct 
 from fprime_openmct.config_server import register_npm_package, install_npm_package, start_npm_package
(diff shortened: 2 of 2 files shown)


src/fprime_gds/executables/run_deployment.py
```diff
@@ -13,9 +13,15 @@
     GdsParser,
     ParserBase,
     StandardPipelineParser,
+    OpenMCTTelemetryPollerParser
 )
 from fprime_gds.executables.utils import AppWrapperException, run_wrapped_application
 
+import fprime_openmct 
+from fprime_openmct.config_server import register_npm_package, install_npm_package, start_npm_package
+from fprime_openmct.src.fprime_openmct.TopAppDictXMLtoOpenMCTJSON import TopologyAppDictionaryJSONifier
```
thomas-bc: We are going to want fprime_openmct to be a soft dependency (i.e. don't error out of `fprime-gds` if `fprime-openmct` is not installed). [Here's an example](https://github.com/fprime-community/fprime-tools/blob/77aa704873e5ad0da1d66b26f7a917cc7842e97d/src/fprime/fpp/visualize.py#L14-L17) of how we do this in fprime-tools, where fprime-visual is a soft dependency.

src/fprime_gds/executables/run_deployment.py
```diff
@@ -148,6 +154,54 @@ def launch_comm(parsed_args):
...
+    """ Convert F-Prime Topology App Dictionary XML to Python Dictionary
+    
+    Args:
+        parsed_args: parsed argument namespace
+    Return:
+        converted OpenMCT JSON Telemetry Definitions
+        Initial States JSON for OpenMCT 
+    """
+
+    openmct_dir = fprime_openmct.__file__.replace('__init__.py', 'src/fprime_openmct/javascript')
```
thomas-bc: I'm not sure I understand why the `openmct_dir` is needed only to be passed as an argument in the openmct functions down below. It may be better to internalize this directly into the fprime-openmct package.

src/fprime_gds/executables/run_deployment.py
```diff
@@ -148,6 +154,54 @@ def launch_comm(parsed_args):
...
+    Return:
+        launched process
+    """
+    openmct_dir = fprime_openmct.__file__.replace('__init__.py', 'src/fprime_openmct/javascript')
+
+
+    pkg = register_npm_package(openmct_dir + "/package.json")
+
+    # Check and install node server if needed
+    install_npm_package(pkg, openmct_dir)
```
thomas-bc: Same as above, we would likely want to internalize this in fprime-openmct

src/fprime_gds/executables/run_deployment.py
```diff
@@ -174,11 +228,22 @@ def main():
...
+
+        #Try polling for telemetry
+        launchers.append(poll_telem)
+
     # Launch launchers and wait for the last app to finish
     try:
         procs = [launcher(parsed_args) for launcher in launchers]
         print("[INFO] F prime is now running. CTRL-C to shutdown all components.")
-        procs[-1].wait()
+        procs[-3].wait()
```
thomas-bc: Why do the last 2 processes need to stay alive? We will likely want to avoid doing that as it impacts all other deployment runs, but if that's really needed for openmct you can just have `if parsed_args.openmct:` that allows for it.

src/fprime_gds/executables/run_deployment.py
mohitsingh999: fixed in the latest commit

src/fprime_gds/executables/run_deployment.py
mohitsingh999: you are right, that is not needed. I have internalized this in my latest commit

src/fprime_gds/executables/run_deployment.py
mohitsingh999: they do not need to stay alive, I reverted back to the original ```procs[-1].wait()```

src/fprime_gds/executables/cli.py
```diff
@@ -782,3 +782,41 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
 
     def handle_arguments(self, args, **kwargs):
         return args
+    
+class OpenMCTTelemetryPollerParser(ParserBase):
```
jwest115, line 786: We may want to include the value of the fprime-gds `--tts-port` argument, since if a different port is used when launching fprime-gds, we get an error in fprime-openmct (because it defaults to 50050). Same probably goes for `--tts-addr`.

src/fprime_gds/executables/cli.py
```diff
@@ -782,3 +782,41 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
...
+    """ Parser for OpenMCT Server arguments  """
+
+    def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
+        """ Return arguments on whether to use OpenMCT, post to a specific URI, and publish telemetry at a specific frequency"""
```
jwest115, line 790: In response to the spell check failure that appears here and in a few other places, unrecognized words can be added to fprime-gds/.github/actions/spelling/expect.txt

src/fprime_gds/executables/run_deployment.py
```diff
@@ -13,9 +13,18 @@
...
+    OpenMCTTelemetryPollerParser
 )
 from fprime_gds.executables.utils import AppWrapperException, run_wrapped_application
 
+# Try to Import FPrime-OpenMCT Python Packages
+try: 
+    from fprime_openmct.config_server import ServerConfig
+    from fprime_openmct.fprime_to_openmct import TopologyAppDictionaryJSONifier
+except:
+    pass
```
thomas-bc: This except should be restricted to only catch import errors, and probably leave a flag behind e.g. `OPENMCT_NOT_INSTALLED=True` that you can check like such: .... (see comment below)

src/fprime_gds/executables/run_deployment.py
```diff
@@ -174,6 +209,20 @@ def main():
     # Launch the desired GUI package
     launchers.append(launch_html)
 
+    # Check if OpenMCT is set to be used. If true, generate OpenMCT States and Launch the OpenMCT Server
+    if parsed_args.openmct:
+
```
thomas-bc, line 218: ... e.g. 
```
if OPENMCT_NOT_INSTALLED: 
     [raise "Error: OpenMCT not installed. Install with....]
```
Otherwise it'll just raise weird errors when trying to use the symbols below.

src/fprime_gds/executables/run_deployment.py
mohitsingh999: should I just have `import openmct` in the try statement then? Something like:
```
try:

     import openmct

except:

     OPENMCT_INSTALLED = False 
```
And then have the ServerConfig and TopologyAppDictionaryJSONifier imports in a conditional statement after?

thomas-bc: Yeah you can actually do

```
try:
    import openmct
except ImportError:
    openmct = None
```

Then you can simply use 

```
if openmct:
     use openmct.ServerConfig() etc...
else:
    raise <error ....>
```

mohitsingh999: Sounds good! I've followed this structure in my most recent commits

src/fprime_gds/executables/cli.py
mohitsingh999: Fixed! I was not parsing the StandardPipeline arguments before and added that in the updated commit

SMorettini:
I'm interested in this feature, what's the state of the pull request?

SMorettini:
I just tested this feature. I like it, is there something I can do to help merge this pull request?

thomas-bc:
@SMorettini we are working on the addition of a plugin architecture to the GDS (https://github.com/nasa/fprime/issues/2268) and we would like the OpenMCT integration to be using it. I believe we decided to hold off on this PR until then. Copying @LeStarch for sanity check and because he's going to be the one doing the plugin architecture.

thomas-bc:
Plugin PR was just merged! https://github.com/fprime-community/fprime-gds/pull/164

We have a plugin system - some work is probably needed to incorporate this into it.

LeStarch:
In order for this to work, we need the "start-up application" plugin defined.  That hasn't been done yet, but isn't much work.

SMorettini:
I don't think this merge request should be merged in this stage. I'm working on some big change on it in my fork. 
This setup is not optimal and doesn't work with big setup with a lot of channels because the  full dictionary for openmct is downloaded every time when a smarter system should be used and only the dictionary for the channel requested you should be downloaded. Moreover it doesn't work except for the channel made by a single value.

ReggieMarr:
Out of curiosity how does openmct compare to [yamcs](https://docs.yamcs.org/python-yamcs-client/mdb/)? My org uses that so I'm looking into a similar effort for that support

timcanham:
@ReggieMarr We are aware of `yamcs,` but haven't gotten around to digging deeply into it. I took a quick look a while ago, and it supports commanding and telemetry aspects that probably could be adapted to F Prime. It doesn't support (as far as I can tell) the  F Prime notion of events.

OpenMCT is really meant to display telemetry. It is great for building views and plots and the like, but it doesn't support commanding and events. Our adaptation was meant to run along side the regular F Prime ground system.

thomas-bc:
I will close this for now as we do not have the traction and capacity required to maintain this for now. Happy to circle back to it if priorities shuffle around in the future.

PR #139, opened 2023-08-18
Update channels to last received with option for latest time

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This adds a setting to control if the latest channel received (by order, or by timestamp) should be displayed.

Defaults to packet order, but users expecting out-of-order packets can adjust in the settings tab.

### DIFF
--- a/src/fprime_gds/flask/static/addons/advanced-settings/addon-templates.js
+++ b/src/fprime_gds/flask/static/addons/advanced-settings/addon-templates.js
@@ -9,6 +9,10 @@ export let advanced_template = `
                 <h3>{{ setting_category.replace("_", " ") }}</h3>
                 <div v-html="settings[setting_category].description"></div>
                 <div class="input-group mb-3" v-for="setting_key in Object.keys(settings[setting_category].settings)">
+                    <small v-if="(settings[setting_category].descriptions || {})[setting_key]">
+                        <strong>{{ setting_key }}</strong>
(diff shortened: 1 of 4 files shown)


LeStarch:
Tested with:
```c++
+++ b/Drv/BlockDriver/BlockDriverImpl.cpp
@@ -2,6 +2,10 @@
 #include <FpConfig.hpp>
 #include <Fw/Types/Assert.hpp>
 
+
+bool well_fake = false;
+Fw::Time fake_time = Fw::Time();
+
 namespace Drv {
 
     BlockDriverImpl::BlockDriverImpl(const char* compName) :
@@ -24,8 +28,16 @@ namespace Drv {
         timer.take();
         // call output timing signal
         this->CycleOut_out(0,timer);
+
+        if (well_fake)
+            fake_time.set(fake_time.getSeconds() - 1, 0);
+        else {
+            well_fake = true;
+            fake_time = this->getTime();
+        }
+
         // increment cycles and write channel
-        this->tlmWrite_BD_Cycles(this->m_cycles);
+        this->tlmWrite_BD_Cycles(this->m_cycles, fake_time);
         this->m_cycles++;
     }
 
```

This causes the cycle count in block driver to count backwards.  When checked, the backward channel updates. When unchecked, the update is not seen.

thomas-bc (approved):
LGTM

PR #141, opened 2023-09-11
Fix scalar input forms

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| fprime_gds/flask |

---

## Change Description

Original change:  _Added the modifier `number` to the scalar argument. This will cast the value to a number and not return a string.((https://v2.vuejs.org/v2/guide/forms.html#number))_  --> This change was breaking the usage of hex, bin and octal because there is not such representation in Javascript.

The new change consists of updating the squash function so any number is actually converted to a number.

## Rationale

This fix is required to have commands containing argument arrays of numbers. Without the Fix, Vue will send to the GDS server an array like `'["1", "2",...]'`. The parsing of the array will fail in Python with the error "Failed to validate all arguments". 

After the fix, the array will be sent to the GDS backend as '[1, 2,...]' and the parsing of the array will work fine.

### DIFF
--- a/src/fprime_gds/flask/static/addons/commanding/arguments.js
+++ b/src/fprime_gds/flask/static/addons/commanding/arguments.js
@@ -8,6 +8,7 @@
  */
 import "../../third-party/js/vue-select.js"
 import {
+    command_bool_argument_template,
     command_enum_argument_template,
     command_array_argument_template,
     command_serializable_argument_template,
@@ -139,6 +140,15 @@ export function squashify_argument(argument) {
     else if (["F64Type", "F32Type"].indexOf(argument.type.name) != -1) {
         value = parseFloat(argument.value);
     }
+    else if (argument.type.name == "BoolType") {
+        if ((typeof(value) === "string") && (["true", "yes"].indexOf(value.toLowerCase()) !== -1)) {
+            value = true;
+        } else if ((typeof(value) === "string") && (["false", "no"].indexOf(value.toLowerCase()) !== -1)) {
+            value = false;
+        } else {
+            console.assert(typeof(value) !== "boolean", "Cannot process boolean, invalid input type")
+        }
+    }
     return value;
 }
 
@@ -246,6 +256,14 @@ Vue.component("command-enum-argument", {
 });
 
 /**
+ * Special boolean processing component to render as a drop-down.
+ */
+Vue.component("command-bool-argument", {
+    ...base_argument_component_properties,
+    template: command_bool_argument_template,
+});
+
+/**
  * Scalar argument processing. Sets up the input type such that numbers can be input with the correct formatting.
  */
 Vue.component("command-scalar-argument", {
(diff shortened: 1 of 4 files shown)


thomas-bc:
Glancing over this, am I understanding correctly that this resolves https://github.com/nasa/fprime/issues/2273 ? Or is that a separate case?

SMorettini:
I didn't see that issue but yes, this pull request resolved exactly that problem.

Tetragramm:
This PR resolves an issue I'm seeing with multi-element parameters.  Not a case noted in either this PR or the linked issue nasa/fprime#2273.
Ex:
```
struct MultiParams {
  param_a: F32,
  param_b: F32
}
```

SMorettini:
For testing my change you could: clone my branch  `https://github.com/SMorettini/fprime-gds/tree/Fix-scalar-input-forms` in your laptop, and then install the customize gds with: `pip install path/to/my/local/gds`. Then if you run `fprime-gds` you should have the one with the fix.

SMorettini:
The previous solution was working but created a new bug, I implemented a new solution that solve the original problem without impacting unsigned integer input.

src/fprime_gds/flask/static/addons/commanding/arguments.js
```diff
@@ -118,6 +118,26 @@ export function squashify_argument(argument) {
             let field = argument.type.MEMBER_LIST[i][0];
             value[field] = squashify_argument(argument.value[field]);
         }
+    } else if (argument.type.name[0] == 'U') {
```
thomas-bc: I know this is done the same way lower in the same file, but it seems to me that this logic of checking the first letter of `type.name` to check the type is flawed. 
Say the argument of your command is a `enum` or struct, type.name will carry the value of the C++ type, which will be `ModuleName::EnumName`. Now say you name the module `Foo`, that gets you `Foo::MyEnum`, so the `argument.type.name[0] == 'F'` condition will be true. So this kind of restricts the usage of module names that start with U, I or F.

@LeStarch I feel like we had noticed this before, does that ring a bell? Or was it just me and I forgot to file an issue?

Anyways, since the bug is already there, it's not necessary to fix it in this PR I would say. I'll create an issue to track it.

thomas-bc: https://github.com/nasa/fprime/issues/2282

LeStarch: This was seen before with Floats and structures/Enums that started with `F`.  The correct fix is to track the exact type names [`I8`, `I16`, ..., `U8`, `U16`] and checking against those.

LeStarch: The below case is arguably "safe" because `inputType` is only used for specific types (scalar, non-enums) as determined by the Vue template. This case is vulnerable to enums because they are not checked in the `if` statement above (only scalars and arrays are).  Thus enums starting with `U`, `I`, or `F` would hit these statements.

It seems we should write a utility function that checks types against a class type.  Something like:

```
is_scalar_type(type, "signed");
``` 

As this risky check has been used multiple times.

LeStarch:
@SMorettini I went ahead and edited this to use full type names to ensure that the code does not inadvertently read types starting with letters. Care to review my one commit?

@thomas-bc care to review?

LeStarch (dismissed):
Awesome change!  I touched it up to ensure a previous bug we've seen has not resurfaced.

src/fprime_gds/flask/static/addons/commanding/arguments.js
LeStarch: Note: declined to write the utility function for now.

SMorettini:
> @SMorettini I went ahead and edited this to use full type names to ensure that the code does not inadvertently read types starting with letters. Care to review my one commit?

Thank you for the change, I didn't have time to make it in the last weeks. For me it seems fine

SMorettini (commented):
Thank you for the change, I didn't have time to make it in the last weeks. 
I cannot approve it because I created the pull request, anyway, I agree with your changes.

LeStarch:
Since I did not see booleans addressed in these if trees, which caused errors, I also fixed: https://github.com/nasa/fprime/issues/2320

thomas-bc (approved):

PR #142, opened 2023-09-13
Added incremental id in history items

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| Event table  |

---
## Change Description

Added an incremental ID to the objects stored in the `HistoryHelper`.

## Rationale

The current implementation works only if there are no events with the same ID sent in the same timestamp. By adding the incremental-id is possible to have a  unique ID for the events. This enables the support of events with the same ID sent in the same instant. 

I know that sending an event with the same id at the same time is very improbable but it happened in my project. We were sending an ERROR in sequence and GDS crashed.

### DIFF
--- a/src/fprime_gds/flask/static/js/datastore.js
+++ b/src/fprime_gds/flask/static/js/datastore.js
@@ -28,6 +28,7 @@ class HistoryHelper {
         this.active_key = active_key;
         this.consumers = [];
         this.active_timeout = null;
+        this.counter = 0;
     }
 
     /**
@@ -57,9 +58,9 @@ class HistoryHelper {
         let _self = this;
         // Break our when no new items returned
         if (new_items.length === 0) { return; }
-        new_items.filter((item) => item.time).forEach((item, index) => {
+        new_items.filter((item) => item.time).forEach((item) => {
             item.datetime = timeToDate(item.time);
-            item.incremental_id = index + this.store.length;
+            item.incremental_id = this.counter++;
         });
         this.consumers.forEach((consumer) => {
             try {
--- a/src/fprime_gds/flask/static/js/datastore.js
+++ b/src/fprime_gds/flask/static/js/datastore.js
@@ -57,8 +57,9 @@ class HistoryHelper {
         let _self = this;
         // Break our when no new items returned
         if (new_items.length === 0) { return; }
-        new_items.filter((item) => item.time).forEach((item) => {
-            item.datetime = timeToDate(item.time)
+        new_items.filter((item) => item.time).forEach((item, index) => {
+            item.datetime = timeToDate(item.time);
+            item.incremental_id = index + this.store.length;
         });
         this.consumers.forEach((consumer) => {
             try {
--- a/src/fprime_gds/flask/static/js/vue-support/event.js
+++ b/src/fprime_gds/flask/static/js/vue-support/event.js
@@ -109,7 +109,7 @@ Vue.component("event-list", {
          * @return {string} unique key
          */
         keyify(item) {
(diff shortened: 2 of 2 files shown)


thomas-bc (commented):
Changes look go to me, but I'm wondering about the motivation behind this. 

> We were sending an ERROR in sequence and GDS crashed.

Does that mean quite literally something like chaining two `log_[...]_MYEVENT()` function calls together? What is prompting this design?

src/fprime_gds/flask/static/js/datastore.js
```diff
@@ -57,8 +57,9 @@ class HistoryHelper {
         let _self = this;
         // Break our when no new items returned
         if (new_items.length === 0) { return; }
-        new_items.filter((item) => item.time).forEach((item) => {
-            item.datetime = timeToDate(item.time)
+        new_items.filter((item) => item.time).forEach((item, index) => {
+            item.datetime = timeToDate(item.time);
+            item.incremental_id = index + this.store.length;
```
LeStarch: Perhaps it would be better to use a GUID or something else appended?  Theoretically, the store can be cleared and thus you could get a repeated incremental ID crossing the "clear" boundary.

Thoughts?

LeStarch: In general, I do appreciate this fix!

LeStarch (changes requested):

SMorettini:
> Does that mean quite literally something like chaining two `log_[...]_MYEVENT()` function calls together? What is prompting this design?

In my code, I had a loop to check for some parameters and send an error for each parameter out of a certain limit. The event was something like `The parameter {id_param} was above the limit`. I used the same event for all the parameters but just changed the `id_param` and in some rare cases, it happened that I sent those events at the same timestamp.  I'll probably change the code to send less recent but still I think GDS should not crash.  

If you want a more detailed example I can try to write down a small example for it.

src/fprime_gds/flask/static/js/datastore.js
SMorettini: I don't understand what you would like to change, is it just the name of the variable or something more?

> Theoretically, the store can be cleared and thus you could get a repeated incremental ID crossing the "clear" boundary.

Yes, this solution is not perfect but it drastically reduce the risk of an error since the id is computed using timestamp+event_id+ incremental id.

thomas-bc:
Ohh ok understood no that makes sense thanks!

src/fprime_gds/flask/static/js/datastore.js
LeStarch: How about this?  Just add a monotonically increasing ID.

```
# Constructor
this.incrementor = 0

# In item handler:
this.incremental_id = _self.incrementor++;
```

This will work across clears of the index.  It is simple and easy to understand.  It will be unique across the events coming in to this specific screen.

A GUID is likely too heavy weight to make sense.

LeStarch:
@SMorettini I made the change here to switch to a basic counter to get incremental ids as this is, I think, less likely to collide due to reused indexes.

We really hope we aren't stepping on your toes touching up your PRs.  We love the work and do not wish to over-burden you with many requests and small changes.  We are also happy to hear your feedback on any of our suggestions/fixes.

On this one, we would like your help to verify the fix, as I cannot seem to duplicate the original behavior.  Hints on how to duplicate it, or testing out the fix would be much appreciated.  I tried many back-to-back event calls, but that did not seem to cause collisions.

Thanks for all you've contributed!

LeStarch (approved):

thomas-bc (approved):

PR #150, opened 2023-11-03
Handle PermissionErrors on downlink

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/1970 |

---
## Change Description

Fixes crashes of the GDS when trying to downlink a file when downlink store is unwritable. Reported https://github.com/nasa/fprime/issues/1970

Tested locally, downlink, change permissions, downlink again. And again... etc. Behaves as expected

### DIFF
--- a/src/fprime_gds/common/files/downlinker.py
+++ b/src/fprime_gds/common/files/downlinker.py
@@ -99,7 +99,10 @@ class FileDownlinker(fprime_gds.common.handlers.DataHandler):
             self.active.open(TransmitFileState.WRITE)
         except PermissionError as exc:
             self.state = FileStates.ERROR
-            LOGGER.warning("Unable to open file for writing. Discarding subsequent downlink packets. " + str(exc))
+            LOGGER.warning(
+                "Unable to open file for writing. Discarding subsequent downlink packets. "
+                + str(exc)
+            )
             return
         LOGGER.addHandler(self.active.log_handler)
         message = "Received START packet with metadata:\n"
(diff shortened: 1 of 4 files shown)


src/fprime_gds/common/files/downlinker.py
```diff
@@ -52,7 +52,6 @@ def __init__(self, directory, timeout=20.0, log_dir=None):
         self.sequence = 0  # Keeps track of what the current sequence ID should be
         self.timer = fprime_gds.common.files.helpers.Timeout()
         self.timer.setup(self.timeout, timeout)
-        os.makedirs(self.__directory, exist_ok=True)
```
thomas-bc, line 55: Already handled here:
https://github.com/fprime-community/fprime-gds/blob/devel/src/fprime_gds/executables/cli.py#L549

src/fprime_gds/common/pipeline/files.py
```diff
@@ -41,6 +42,11 @@ def setup_file_handling(
         )
         file_decoder.register(self.__downlinker)
         distributor.register("FW_PACKET_HAND", self.__uplinker)
+        if not os.access(down_store, os.W_OK):
```
thomas-bc, line 46: This warning will be thrown when the `down_store` already exists but is not writeable. If down_store doesn't exist **and** cannot be created, GDS fails to start up with PermissionDenied to create down_store. Should it have the same behavior in both cases?

LeStarch: Fail instead of warning here.

LeStarch (changes requested):
One minor fix

LeStarch (approved):

PR #152, opened 2023-11-22
Remove flask_uploads.py

### LINKED ISSUE #2352: Remove `flask_uploads.py` from the GDS
| | |
|:---|:---|
|**_F´ Version_**| 3.3.x |
|**_Affected Component_**|  |
---
## Feature Description

[flask_uploads.py](https://github.com/fprime-community/fprime-gds/blob/devel/src/fprime_gds/flask/flask_uploads.py) is an old script that was forked from Flask, allowing to upload files. It shouldn't be necessary to keep it anymore. 
We should remove it, and keep only the minimal set of functionality that fits the use case.
Preliminary work has been done here: https://github.com/fprime-community/fprime-gds/tree/discard-flask-uploads
I believe this is working - just need to confirm, test and polish the branch before opening a PR.

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2352 |

---
## Change Description

Fixes https://github.com/nasa/fprime/issues/2352

Removes the need for the flask_uploads.py module.

Additionally adds in a `--remote-sequence-directory` option to control where sequence binaries are uploaded on the remote (default `/seq`, which is often not writeable when testing locally).


thomas-bc:
Not sure what's up with the Unit tests, they pass locally... 

No need to merge this prior to v3.4 release

src/fprime_gds/common/pipeline/files.py
```diff
@@ -43,10 +43,11 @@ def setup_file_handling(
         )
         file_decoder.register(self.__downlinker)
         distributor.register("FW_PACKET_HAND", self.__uplinker)
-        if not os.access(down_store, os.W_OK):
+        try:
+            Path(down_store).mkdir(parents=True, exist_ok=True)
+        except PermissionError:
```
thomas-bc, line 48: Changing the way we check for permissions here because `os.access(down_store, os.W_OK)` returns False if down_store doesn't exist, even if user could create down_store and have permissions on it. Which is really what we're trying to test here. This is already handled in StandardPipeline.setup(), but trying to make this test more robust / agnostic to StandardPipeline.

src/fprime_gds/executables/cli.py
```diff
@@ -535,18 +538,26 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
...
+                "required": False,
+                "type": str,
+                "help": "Directory to save command sequence binaries, on the remote FSW. Default: %(default)s",
+            },
         }
 
     def handle_arguments(self, args, **kwargs):
         """Handle arguments as parsed"""
-        os.makedirs(args.files_directory, exist_ok=True)
+        os.makedirs(args.files_storage_directory, exist_ok=True)
```
LeStarch: Do the permissions check here.

src/fprime_gds/flask/updown.py
```diff
@@ -112,6 +113,53 @@ def post(self):
...
+
+        :param file_storage: The uploaded file to save.
+        """
+        if not isinstance(file_storage, FileStorage):
+            raise TypeError("file_storage must be a werkzeug.FileStorage")
+
+        filename = Path(secure_filename(file_storage.filename)).name
+        dest_dir = Path(self.dest_dir)
+
+        # not necessary
```
LeStarch: Add permissions check and is necessary, to keep independent of StandardPipeline

LeStarch (commented):
Two comment

src/fprime_gds/executables/cli.py
thomas-bc: Fixed

src/fprime_gds/flask/updown.py
thomas-bc: Fixed

LeStarch (approved):

PR #153, opened 2023-12-07
Add pyproject.toml

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/1995 |

---
## Change Description

- Add `pyproject.toml` 
- Strip setup.py from old config
- Update publishing workflow to use trusted publishers mechanism

## Rationale

Conform to PEP517, 518 et al. Modernize our Python setup so as not to be left behind. Potentially help with Python 3.12 support

TODO:
- sanity check that this switch won't break other external things?
- alpha release and let CI fly


pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Operating System :: Unix",
+    "Operating System :: POSIX",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
LeStarch, line 27: Should we add Py 12 back?

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+####
+[project.scripts]
+fprime-cli = "fprime_gds.executables.fprime_cli:main"
+fprime-seqgen = "fprime_gds.common.tools.seqgen:main"
+
+[project.gui-scripts]
+fprime-gds = "fprime_gds.executables.run_deployment:main"
+
+# For Pytest fixtures
+[project.entry-points."pytest11"]
```
LeStarch, line 66: Why is pytest11 in quotes here?

LeStarch (commented):
Minor questions.

pyproject.toml
thomas-bc: Yes let's! will add to other PRs as well

pyproject.toml
thomas-bc: It seems to be the notation https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins

I'm guessing it's because it's user-defined (in the plugin) and not part of the standard template or some other terminology

LeStarch (approved):

PR #155, opened 2023-12-11
Fix/enum representation types

JackNWhite:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Jack White |
|**_Affected Component_**| XML loader |
|**_Related Issue(s)_**| fprime-community/fprime-tools/issues/17 |

---
## Change Description

When loading enums, XML loader checks whether this field exists for the enum in question, and defaults to I32.

## Rationale

Serialization/representation types are part of the F' and FPP specs, but are not, in fact, supported.

## Accompanying PRs
- fprime-community/fprime-tools/pull/184
- nasa/fprime/pull/2419


src/fprime_gds/common/loaders/xml_loader.py
```diff
@@ -218,14 +219,20 @@ def get_enum_type(self, enum_name, xml_obj):
         for enum in enum_section:
             # Check enum name
             if enum.get(self.ENUM_TYPE_TAG) == enum_name:
+                # Get serialize/representation type, if present
+                serialize_type = enum.get(self.ENUM_SERIALIZE_TYPE_TAG)
```
LeStarch: Can you add a default in this call? e.g.:

```
serialize_type = enum.get(self.ENUM_SERIALIZE_TYPE_TAG, "I32")
```

I don't remember if this will work here or not, but most python get calls support this.

LeStarch:
This looks good too.  We'll need to fix the test.  However, we'll need an alpha release of `fprime-tools` and to add that to the setup.py file as we've changed the underlying interface.

src/fprime_gds/common/loaders/xml_loader.py
JackNWhite: This seems to work OK.

LeStarch:
This is approved once we get an alpha release of the fprime-tools fix.

LeStarch:
Closed And reopened to trick CI

LeStarch (approved):

PR #164, opened 2024-03-07
Initial GDS plugin implementation

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Initial plugin integration.

## Rationale

Plugins for the people!

See UTs.

## Future Work

Call out other plugins!

### DIFF
--- a/src/fprime_gds/common/communication/framing.py
+++ b/src/fprime_gds/common/communication/framing.py
@@ -17,6 +17,7 @@ import struct
 import sys
 
 from .checksum import calculate_checksum
+from fprime_gds.plugin.definitions import gds_plugin_implementation
 
 
 class FramerDeframer(abc.ABC):
@@ -191,6 +192,13 @@ class FpFramerDeframer(FramerDeframer):
             # Case of not enough data for a full packet, return hoping for more later
             return None, data, discarded
         return None, data, discarded
+
+    @classmethod
+    @gds_plugin_implementation
+    def register_framing_plugin(cls):
+        """ Register a bad plugin """
+        return cls
+
 
 
 class TcpServerFramerDeframer(FramerDeframer):
--- a/src/fprime_gds/common/communication/framing.py
+++ b/src/fprime_gds/common/communication/framing.py
@@ -16,7 +16,7 @@ import copy
 import struct
 import sys
 
-from .checksum import calculate_checksum
+from .checksum import calculate_checksum, CHECKSUM_MAPPING
 from fprime_gds.plugin.definitions import gds_plugin_implementation
 
 
@@ -98,10 +98,11 @@ class FpFramerDeframer(FramerDeframer):
     HEADER_FORMAT = None
     START_TOKEN = None
 
-    def __init__(self):
+    def __init__(self, checksum_type):
         """Sets constants on construction."""
         # Setup the constants as soon as possible.
         FpFramerDeframer.set_constants()
+        self.checksum = checksum_type
 
     @classmethod
     def set_constants(cls):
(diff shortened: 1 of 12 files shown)


src/fprime_gds/common/communication/framing.py
```diff
@@ -192,6 +193,13 @@ def deframe(self, data, no_copy=False):
             return None, data, discarded
         return None, data, discarded
 
+    @classmethod
+    @gds_plugin_implementation
+    def register_framing_plugin(cls):
+        """ Register a bad plugin """
```
thomas-bc, line 222: probably want to change the docstring here :p

src/fprime_gds/plugin/system.py
```diff
@@ -0,0 +1,119 @@
...
+
+PROJECT_NAME = definitions.PROJECT_NAME
+LOGGER = logging.getLogger(__name__)
+
+_NAME_REGEX = re.compile(r"^register_(\w+)_plugin")
+_TYPE_MAPPINGS = {
+    definitions.register_framing_plugin: framing.FramerDeframer
+}
+_SUPPLIED_PLUGIN_MODULES_OR_CLASSES = [
+    framing.FpFramerDeframer
```
thomas-bc: Can this not be inferred from the set of values in _TYPE_MAPPINGS ? Or is there a difference? Might be easier to maintain as the number of pluginable classes grow.

src/fprime_gds/executables/cli.py
```diff
@@ -805,3 +808,64 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
...
+            } for category in Plugins.system().get_categories()
+        }
+
+    def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
+        """ Return arguments to retrieve channels/events/commands in specific ways """
+
+        arguments: Dict[Tuple[str, ...], Dict[str, Any]] = {}
+        for category, selections in self._plugin_map.items():
+            arguments.update({
+                (f"--{category}-selection",): {
```
thomas-bc: Do we want to allow category names to grow past a single word? If so, should we do something like `category.replaceAll('_', '-')` here?

src/fprime_gds/executables/cli.py
```diff
@@ -805,3 +808,64 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
...
+    def __init__(self):
+        """ Initialize the plugin information for this parser """
+        self._plugin_map = {
+            category: {
+                self.get_selection_name(selection): selection for selection in Plugins.system().get_selections(category)
+            } for category in Plugins.system().get_categories()
+        }
+
+    def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
+        """ Return arguments to retrieve channels/events/commands in specific ways """
```
thomas-bc: Could remove/rewrite the docstring here

LeStarch:
@thomas-bc  this is done!

src/fprime_gds/plugin/system.py
thomas-bc: Nevermind the latest commit clarified the difference

src/fprime_gds/executables/cli.py
LeStarch: I do not want to do that now.  Since our team is the one that adds plugin categories, we can add it if we ever get a category with multiple words.

src/fprime_gds/common/communication/framing.py
LeStarch: Since been fixed.

src/fprime_gds/plugin/system.py
LeStarch: One could ask each type for subclasses and then use those.  However, this has one confusing blind-spot:  did that class get imported?  If not, it will not show up.  This is really hard to understand and debug.

Rather, I opted for hand-curated configuration.  That way when you plugin does not show up, a search through that file reveals each step you need to do without some implied magic.

> Remember, this only affects "built-in" plugin selections added by our team.  Any external supplied plugin should publish their work as a python module with a pluggy entrypoint....which will be automatic.

src/fprime_gds/executables/cli.py
LeStarch: Fixed.

thomas-bc (approved):

PR #173, opened 2024-04-23
JSON Dictionary Integration

### LINKED ISSUE #2592: Integrate JSON dictionary in the GDS
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

JSON Dictionary is here: https://github.com/nasa/fprime/issues/2263

In the GDS, we have a whole system to ingest the XML dictionary and make that data available through data types. We need to ingest the JSON dictionary instead. 
Retiring the old data types and using the JSON data structure itself sounds like a reasonable design... will need to look into

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2592 |

---
## Change Description

Implements https://github.com/nasa/fprime/issues/2592

### DIFF
--- a/src/fprime_gds/common/loaders/ch_json_loader.py
+++ b/src/fprime_gds/common/loaders/ch_json_loader.py
@@ -181,21 +181,29 @@ class ChJsonLoader:
                 type_name,
                 qualified_type.get("identifiers"),
                 qualified_type.get("representationType").get("name"),
-                # self.parse_type(qualified_type.get("representationType")),
             )
 
         if qualified_type.get("kind") == "struct":
-            struct_member_list = [
-                (
-                    name,
-                    self.parse_type(member_dict.get("type")),
-                    member_dict.get("type", {}).get("format", "%s"),
-                    member_dict.get("type", {}).get("description", ""),
-                )
-                for name, member_dict in qualified_type.get("members", {}).items()
-            ]
+            struct_members = []
+            for name, member_dict in qualified_type.get("members").items():
+                member_type_dict = member_dict.get("type")
+                member_type_obj = self.parse_type(member_type_dict)
+
+                # For member arrays (declared inline, so we create a type on the fly)
+                if member_dict.get("size") is not None:
+                    member_type_obj = ArrayType.construct_type(
+                        f"Array_{member_type_obj.__name__}_{member_dict.get('size')}",
+                        member_type_obj,
+                        member_dict.get("size"),
+                        member_dict.get("type").get("format", "%s"),
+                    )
+
+                fmt_str = member_type_dict.get("format", "%s")
+                description = member_type_dict.get("description", "")
+                struct_members.append((name, member_type_obj, fmt_str, description))
+
             return SerializableType.construct_type(
                 type_name,
-                struct_member_list,
+                struct_members,
             )
 
--- a/src/fprime_gds/common/loaders/ch_json_loader.py
+++ b/src/fprime_gds/common/loaders/ch_json_loader.py
@@ -8,10 +8,13 @@ Loads flight dictionary (JSON) and returns id and mnemonic based Python dictiona
 
 from fprime_gds.common.templates.ch_template import ChTemplate
 from fprime_gds.common.loaders.json_loader import JsonLoader
+from fprime_gds.common.data_types.exceptions import GdsDictionaryParsingException
 
 
 class ChJsonLoader(JsonLoader):
     """Class to load python based telemetry channel dictionaries"""
+
+    CHANNELS_FIELD = "telemetryChannels"
 
     ID = "id"
     NAME = "name"
@@ -40,7 +43,12 @@ class ChJsonLoader(JsonLoader):
         id_dict = {}
         name_dict = {}
 
-        for ch_dict in self.json_dict["telemetryChannels"]:
+        if self.CHANNELS_FIELD not in self.json_dict:
+            raise GdsDictionaryParsingException(
+                f"Ground Dictionary missing '{self.CHANNELS_FIELD}' field: {str(self.json_file)}"
+            )
+
+        for ch_dict in self.json_dict[self.CHANNELS_FIELD]:
             # Create a channel template object
             ch_temp = self.construct_template_from_dict(ch_dict)
 
@@ -54,10 +62,23 @@ class ChJsonLoader(JsonLoader):
         )
 
     def construct_template_from_dict(self, channel_dict: dict) -> ChTemplate:
-        component_name = channel_dict[self.NAME].split(".")[0]
-        channel_name = channel_dict[self.NAME].split(".")[1]
-        channel_type = channel_dict[self.TYPE]
-        type_obj = self.parse_type(channel_type)
+        try:
+            ch_id = channel_dict[self.ID]
+            # The below assignment also raises a ValueError if the name does not contain a '.'
+            component_name, channel_name = channel_dict[self.NAME].split(".")
+            if not component_name or not channel_name:
+                raise ValueError()
+
+            type_obj = self.parse_type(channel_dict[self.TYPE])
+        except ValueError as e:
+            raise GdsDictionaryParsingException(
+                f"Channel dictionary entry malformed, expected name of the form '<COMP_NAME>.<CH_NAME>' in : {str(channel_dict)}"
+            )
+        except KeyError as e:
+            raise GdsDictionaryParsingException(
+                f"{str(e)} key missing from Channel dictionary entry or its associated type in the dictionary: {str(channel_dict)}"
+            )
+
         format_str = JsonLoader.preprocess_format_str(channel_dict.get(self.FMT_STR))
 
         limit_field = channel_dict.get(self.LIMIT_FIELD)
@@ -71,7 +92,7 @@ class ChJsonLoader(JsonLoader):
         limit_high_orange = limit_high.get(self.LIMIT_ORANGE) if limit_high else None
 
         return ChTemplate(
-            channel_dict[self.ID],
+            ch_id,
             channel_name,
             component_name,
             type_obj,
--- a/src/fprime_gds/common/loaders/ch_json_loader.py
+++ b/src/fprime_gds/common/loaders/ch_json_loader.py
@@ -49,19 +49,19 @@ class ChJsonLoader(JsonLoader):
             # Create a channel template object
             ch_temp = self.construct_template_from_dict(ch_dict)
 
-            id_dict[ch_dict[self.ID_FIELD]] = ch_temp
-            name_dict[ch_dict[self.NAME_FIELD]] = ch_temp
+            id_dict[ch_temp.get_id()] = ch_temp
+            name_dict[ch_temp.get_full_name()] = ch_temp
 
         return (
-            dict(sorted(id_dict.items())),
-            dict(sorted(name_dict.items())),
-            ("unknown", "unknown"),
+            id_dict,
+            name_dict,
+            self.get_versions(),
         )
 
-    def construct_template_from_dict(self, channel_dict: dict):
+    def construct_template_from_dict(self, channel_dict: dict) -> ChTemplate:
         component_name = channel_dict[self.NAME_FIELD].split(".")[0]
         channel_name = channel_dict[self.NAME_FIELD].split(".")[1]
-        channel_type = channel_dict.get("type")
+        channel_type = channel_dict["type"]
         type_obj = self.parse_type(channel_type)
         format_str = JsonLoader.preprocess_format_str(
             channel_dict.get(self.FMT_STR_FIELD)
--- a/src/fprime_gds/common/loaders/ch_json_loader.py
+++ b/src/fprime_gds/common/loaders/ch_json_loader.py
@@ -8,18 +8,17 @@ Loads flight dictionary (JSON) and returns id and mnemonic based Python dictiona
 
 from fprime_gds.common.templates.ch_template import ChTemplate
 from fprime_gds.common.loaders.json_loader import JsonLoader
+from enum import Enum
 
 
 class ChJsonLoader(JsonLoader):
     """Class to load python based telemetry channel dictionaries"""
 
-    # Field names in the python module files (used to construct dictionaries)
-    ID_FIELD = "id"
-    NAME_FIELD = "name"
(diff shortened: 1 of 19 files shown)


test/fprime_gds/common/utils/test_string_util.py
```diff
@@ -7,56 +7,63 @@
...
         self.assertEqual(expected, actual)
 
     def test_format_bad_case(self):
         template = "Opcode 0x%04X dispatched to port %04d and value %0.02f"
         values = ("181", "8", "0.123")
         with self.assertRaises(ValueError):
-            self.assertEqual(format_string_template(template, values))
+            self.assertEqual(
+                format_string_template(preprocess_c_style_format_str(template), values)
+            )
```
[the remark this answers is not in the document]
thomas-bc: Fixed in test code

test/fprime_gds/common/utils/test_string_util.py
```diff
@@ -65,63 +72,137 @@
...
+        actual = format_string_template(preprocess_fpp_format_str(template), values)
+        self.assertEqual(expected, actual)
+
+    def test_fpp_format_bad_case(self):
+        template = "Opcode 0x{04X} dispatched to port {04d} and value {0.02f}"
+        values = ("181", "8", "0.123")
+        with self.assertRaises(ValueError):
+            self.assertEqual(
+                format_string_template(preprocess_fpp_format_str(template), values)
+            )
```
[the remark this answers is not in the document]
thomas-bc: Fixed in test code

src/fprime_gds/flask/static/addons/commanding/command-input.js
```diff
@@ -83,7 +83,8 @@ Vue.component("command-input", {
         this.$root.$refs.command_input = this;
     },
     data: function() {
-        let selected = command_assignment_helper(null, [], "CMD_NO_OP");
+        // Select CMD_NO_OP by default if it exists, otherwise select the first command
+        let selected = command_assignment_helper("cmdDisp.CMD_NO_OP", [], "CMD_NO_OP");
```
LeStarch, line 87: What does this change do?

src/fprime_gds/executables/utils.py
```diff
@@ -192,21 +193,31 @@ def find_dict(root: Path) -> Path:
...
         for child in dict_dir.iterdir()
         if child.is_file() and child.name.endswith("Dictionary.xml")
     ]
+    json_dicts = [
+        child
+        for child in dict_dir.iterdir()
+        if child.is_file() and child.name.endswith("Dictionary.json")
+    ]
+    # Select json dictionary if available, otherwise use xml dictionary
+    dicts = json_dicts or xml_dicts
```
LeStarch: What is this?  It seems this should result in a bool, yet `len(dicts)` is used below.  Perhaps:

```
dicts = json_dicts if json_dicts else xml_dicts
```

Is clearer?  Or is `list or list` a pythonism?

src/fprime_gds/common/loaders/ch_json_loader.py
```diff
@@ -0,0 +1,86 @@
...
+            name_dict[ch_temp.get_full_name()] = ch_temp
+
+        return (
+            id_dict,
+            name_dict,
+            self.get_versions(),
+        )
+
+    def construct_template_from_dict(self, channel_dict: dict) -> ChTemplate:
+        component_name = channel_dict[self.NAME].split(".")[0]
```
LeStarch: why not `component_name, channel_name = channel_dict[self.name].split(".")`.   

Also note, this will break on malformed data (e.g. the split only has one item).  Probably should trap that somehow.

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+from fprime.common.models.serialize.serializable_type import SerializableType
+from fprime.common.models.serialize.string_type import StringType
+from fprime.common.models.serialize.type_base import BaseType
+
+from fprime_gds.common.utils.string_util import preprocess_fpp_format_str
+from fprime_gds.common.loaders import dict_loader
+
+
+PRIMITIVE_TYPE_MAP = {
+    "I8": numerical_types.I8Type,
```
LeStarch, line 26: Does this not already exist somewhere in fprime.common.models?  I don't think we should reinvent the wheel.

Note: I looked for it and did not find it.  So perhaps this is ok for now?  It would be nice to have in `fprime-tools` a translator from short name to type and back again.... but not in this PR.

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def get_versions(self):
+        """
+        Get the framework and project versions of the dictionary
+
+        Returns:
+            A tuple of the framework and project versions
+        """
+        if "metadata" not in self.json_dict:
+            # REVIEW NOTE: Should we raise an error instead?
```
LeStarch: It probably makes sense since this is in the defined spec to raise an error in these two cases:

1. metadata not found
2. frameworkVersion not found

As someday we may have hard-requirements on these items existing.  Not a strong opinion.

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+        if type_name is None:
+            raise ValueError(
+                f"Channel entry in dictionary has no `name` field: {str(type_dict)}"
+            )
+
+        if type_name in PRIMITIVE_TYPE_MAP:
+            return PRIMITIVE_TYPE_MAP[type_name]
+
+        if type_name == "string":
+            # REVIEW NOTE: All strings of same size will share the same type. I believe this is good?
```
LeStarch: Yes, this is good.  The defining properties of these strings are "name" and "max length" and as such, I see no reason to define multiple types.

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+            return self.parsed_types[type_name]
+
+        # Parse new enum/array/serializable types
+        qualified_type = None
+        for type_def in self.json_dict.get("typeDefinitions", []):
+            if type_name == type_def.get("qualifiedName"):
+                qualified_type = type_def
+                break
+
+        if qualified_type is None:
```
LeStarch: Could be written as a `for else` which would make it's search-raise-if-not-found behavior a bit easier to follow

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+            EnumType: The constructed EnumType object.
+
+        """
+        enum_dict = {}
+        for member in qualified_type.get("enumeratedConstants"):
+            enum_dict[member.get("name")] = member.get("value")
+        enum_type = EnumType.construct_type(
+            type_name,
+            enum_dict,
+            qualified_type.get("representationType").get("name"),
```
LeStarch: Do you have a trap for `KeyError` somewhere?  `KeyError` should translate to `ValueError("Malformed type definition")`

LeStarch: A minor nit, when calling `get` without a default, I much prefer `item["syntax"]` as it makes it blatant that there is a KeyError possibility.

Thus:
1. Item must exist, or raise: dict[item]
2. Item may exist, or handle: dict(item, some_default)

Personal preference, but it helps me quickly understand the exception paths in code like this.

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+
+            # For member arrays (declared inline, so we create a type on the fly)
+            if member_dict.get("size") is not None:
+                member_type_obj = ArrayType.construct_type(
+                    f"Array_{member_type_obj.__name__}_{member_dict.get('size')}",
+                    member_type_obj,
+                    member_dict.get("size"),
+                    member_dict.get("type").get("format", "{}"),
+                )
+            # TODO: does this need to be preprocessed or something?
```
LeStarch: Probably?  Seems like you have a FPP format string parser function, so it should be called on it, yes?

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+            member_type_dict = member_dict.get("type")
+            member_type_obj = self.parse_type(member_type_dict)
+
+            # For member arrays (declared inline, so we create a type on the fly)
+            if member_dict.get("size") is not None:
+                member_type_obj = ArrayType.construct_type(
+                    f"Array_{member_type_obj.__name__}_{member_dict.get('size')}",
+                    member_type_obj,
+                    member_dict.get("size"),
+                    member_dict.get("type").get("format", "{}"),
```
LeStarch: Preprocess format?

src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -0,0 +1,218 @@
...
+
+        Returns:
+            ArrayType: The constructed ArrayType object.
+
+        """
+        array_type = ArrayType.construct_type(
+            type_name,
+            self.parse_type(qualified_type.get("elementType")),
+            qualified_type.get("size"),
+            qualified_type.get("elementType").get("format", "{}"),
```
LeStarch: Preprocess format?

LeStarch (changes requested):
First, let us be clear, this is top-rate work.  It is also an implementation of a fairly critical subsystem.  Thus I made a lot of comments.  There is one thing I'd like you to think about: Error checking and resolution.

I want there to be a clear story on what happens and what is returned on malformed JSON data.  Given what I see, it is probably a ValueError, or a custom MalformedDictionary error, or a combination.  All other errors should be down-converted to a known set of expressive errors.  

This will help a lot in determining what to do in the layer above and avoid the random exception problem.

Let me know if you have questions.

src/fprime_gds/executables/utils.py
thomas-bc: Your proposed syntax is much clearer, I will switch to it.
The effect is exactly the same -- I believe `list or list` is a pythonism (arguably a bad one) to return the first non-empty list (`[]` if all empty)

src/fprime_gds/flask/static/addons/commanding/command-input.js
thomas-bc: Without it, in the `command_assignment_helper()` function we match both `CMD_NO_OP` and `CMD_NO_OP_STRING`. 
Because commands are not ordered in the JSON dictionary, it happens that CMD_NO_OP_STRING appears first (at least in `Ref/`), and therefore the default command that is selected when the GDS opens is `CMD_NO_OP_STRING`. 

I wanted to keep `CMD_NO_OP` first, so instead of ordering the dictionary which sounded like overkill, I found this little trick.

The behavior is then
```
if `cmdDisp.CMD_NO_OP` exists, return
elif commands match `CMD_NO_OP`, return first match
else return first command in dictionary
```

LeStarch: Right.  However, this does not work for projects whose command dispatcher is not named `cmdDisp`.  In that case, the weird behavior will result.

This does yield several questions:

1. Is the drop-down sorted alphabetically?  If not, it should be.
2. Does typing in `CMD_NO_OP` yield an improperly sorted list?  If so, we probably should fix that too.

src/fprime_gds/executables/utils.py
LeStarch: Cool, glad we agree.  I am not sure I like this particular pythonism, but the pattern is used a lot in javascript....

src/fprime_gds/flask/static/addons/commanding/command-input.js
thomas-bc: The drop-down _is_ sorted correctly, but that's because it sorts them when it retrieves them ([here](https://github.com/nasa/fprime-gds/blob/devel/src/fprime_gds/flask/static/addons/commanding/command-input.js#L211-L225)) from the datastore.
The commands in the datastore are not sorted because it comes straight from the dictionary, which is not sorted in the JSON implementation. So when retrieving the list of matches with `command_assignment_helper()`, it is not sorted correctly. Maybe it should be the responsibility of the JSON loader to sort the dictionaries?

LeStarch: Is it worth sorting the dictionary at load time to offset this minor discrepancy (custom cmdDisp name and getting NO_OP_STRING instead of NO_OP)?  Otherwise I see no need to.

LeStarch: I think we should fix this issue (either sorting at load time, or sorting before running this search) because:

1. Users will expect it to work consistently even with custom component names
2. It will take more work to document and explain the discrepancy than it will be to fix it
3. Confusion will cost us loyalty (even if it is an iota of loyalty) and as such we should document or fix.

src/fprime_gds/common/loaders/json_loader.py
thomas-bc: Done. I don't use for-else much because it's very python-specific and not always intuitive what it's doing. Especially here where it could look like an indentation error.

src/fprime_gds/common/loaders/json_loader.py
thomas-bc: I have not been able to find it either

src/fprime_gds/common/loaders/json_loader.py
thomas-bc: Refactored with the `dict[item]` vs `dict.get(item)` and added error handling

src/fprime_gds/common/loaders/ch_json_loader.py
thomas-bc: Refactored and trapped

src/fprime_gds/flask/static/addons/commanding/command-input.js
thomas-bc: I have added a fix to sort the dictionaries at load time

Ideally they'd be sorted from the output of `fpp-to-dict`... but also we shouldn't be relying on it here. I'm happy to keep it like this, let me know.

LeStarch (approved):

PR #174, opened 2024-05-01
Default to ZMQ with --no-zmq to fallback on TTS

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2520 |

---
## Change Description

Defaults to using the ZMQ transport layer, falling back to TTS by user request with `--no-zmq`

## Rationale

https://github.com/nasa/fprime/issues/2520

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -184,7 +184,7 @@ class ParserBase(ABC):
             sys.exit(-1)
         except Exception as exc:
             print(f"[ERROR] {exc}", file=sys.stderr)
-            raise
+            sys.exit(-1)
         return args_ns, parser
 
     @staticmethod
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -569,7 +569,7 @@ class MiddleWareParser(ParserBase):
                 "dest": "zmq",
                 "action": "store_false",
                 "help": "Disable ZMQ transportation layer, falling back to TCP socket server.",
-                "default": False,
+                "default": True,
             },
             ("--zmq-transport",): {
                 "dest": "zmq_transport",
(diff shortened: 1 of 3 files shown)


src/fprime_gds/executables/cli.py
```diff
@@ -198,7 +185,6 @@ def parse_args(
         except Exception as exc:
             print(f"[ERROR] {exc}", file=sys.stderr)
             raise
-            sys.exit(-1)
```
LeStarch, line 201: Either we should exist here with the above printed error, or we should "raise Exception(f"[ERROR] {exc}")" and let it print elsewhere, or remove the except block altogether.

LeStarch (changes requested):
We must clean-up some ugly error handling.  Note: I probably introduced the raise unintentionally.

Apart from that there is on broad-stroke to consider:  this now requires `zmq` as a package, which is a different disposition than "use zmq if installed".  It is likely advisable to to make this change, to simplify our support-space, but I want a moment to think through this while the other issue is touched-up.

src/fprime_gds/executables/cli.py
thomas-bc: I removed the optional nature of the ZMQ dependency in the code because we have removed it from the optional dependency of the Python package. It is now a formal dependency in [pyproject.toml](https://github.com/nasa/fprime-gds/blob/devel/pyproject.toml#L36) and part of https://github.com/nasa/fprime/blob/devel/requirements.txt

LeStarch: Makes sense!

thomas-bc: Reverted the call back to sys.exit(), as parse_args() isn't always in a try-catch block across our code base. It's also handled the same way right above.

LeStarch (approved):

PR #177, opened 2024-07-16
Seqgen use json instead of xml

Lex-ari:
| | |
|:---|:---|
|**_Related Issue(s)_**| #173 |

---
## Change Description

This makes fprime-seqgen uses json dictionaries instead of xml for sequencing.
Small typo fixes.

## Rationale

Current implementation breaks when submitting a json either through the terminal or through GDS. This small fix properly converts them into binaries (and GDS properly uploads to spacecraft).

@Joshua-Anderson @thomas-bc @timcanham


Joshua-Anderson (commented):
🔥 This is awesome - did you do any testing to validate this works?

I expected that there would be some unit tests but [the only tests I saw were empty](https://github.com/nasa/fprime-gds/blob/devel/test/fprime_gds/common/tools/seqgen_unit_test.py) If you have some time writing a unit test for fprime-seqgen would probably be super helpful.

You may want to update the command line arguments helptext to specify that you need to supply a JSON dictionary file.

This will also require updating the fprime integration tests [here](https://github.com/nasa/fprime/blob/devel/ci/tests/fputil.bash#L62)

Joshua-Anderson:
@thomas-bc could you kick off unit tests to see if this broke any unit tests that we're not aware of?

Lex-ari:
> 🔥 This is awesome - did you do any testing to validate this works?

Tested on a sample deployment that ran a couple of CMD_NO_OPs. I used [this project](https://github.com/fprime-community/fprime-hub-pattern-example) and this [simple_sequence.seq.txt](https://github.com/user-attachments/files/16254848/simple_sequence.seq.txt) (without the .txt).

I could make the test case use a dictionary.json from a fresh deployment and generate a sequence binary off from it, but it might be a "circular" self-validating test. Any suggestions?

Joshua-Anderson:
Yeah, looking at the code structure the only straightforward way to test without refactoring is to provide a simple dictionary and input sequence and make sure it matches an expected output sequence binary.

It feels a bit a circular, but does serve as a good regression test that python and other GDS infrastructure updates don't break the sequence compilation process.

LeStarch:
@thomas-bc do we have tests for the JSON loader?

LeStarch (dismissed):
LGTM.  I'll let you and @Joshua-Anderson determine if tests belong in this PR or another.

test/fprime_gds/common/tools/seqgen_unit_test.py
```diff
@@ -1,7 +1,42 @@
...
+    # Passing these executions is a test failure.
+    @unittest.expectedFailure 
+    def test_fail_unmatched_command_sequence(self):
+        self.diff_input_sequence(
+            "./input/simple_bad_sequence.seq",
+            "./expected/simple_sequence_expected.bin",
+            "./resources/simple_dictionary.json"
+        )
+        
+    def diff_input_sequence(self, input_sequence, expected_output, dictionary):
```
Joshua-Anderson: This method name might be a little misleading, since you're proving an input sequence but diffing against the expected output sequence binary. Maybe consider renaming

test/fprime_gds/common/tools/seqgen_unit_test.py
```diff
@@ -1,7 +1,42 @@
...
+        self.assertEqual(self.diff_input_sequence(
+            "./input/simple_sequence.seq",
+            "./expected/simple_sequence_expected.bin",
+            "./resources/simple_dictionary.json"
+        ), True)
+    
+    # This test is expected to fail due to unmatched command in
+    # simple_bad_sequence.seq and seqgen should produce a SeqGenException. 
+    # Passing these executions is a test failure.
+    @unittest.expectedFailure
```
Joshua-Anderson: Instead of marking that you expect a test case to fail, I usually prefer using [`assertRaises`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) to check that the method raises the expected assertion and isn't just generally crashing.

Joshua-Anderson (approved):
🚀 Looks good to me, thanks @Lex-ari!

Make sure to fix the spellcheck errors

LeStarch (approved):

PR #181, opened 2024-11-07
Fpy Bytecode Serialization Tool

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| [nasa/fprime#2919](https://github.com/nasa/fprime/issues/2919) |

---
This PR adds a simple tool to take Fpy bytecode in a human-readable form, and convert it into the binary form that the `FpySequencer` understands.

As a reminder, this is not intended to be the natural, high-level interface to the Fpy sequencing language. Instead, this is a tool to jumpstart the development of the sequencer and for power-users who do not want to compile from Fpy.


zimri-leisher:
Looks like this is failing to run some builds because I'm using a Python feature from 3.10, type subscripting. I will probably just include 
`from __future__ import annotations` unless you'd rather I remove this syntax entirely.

Joshua-Anderson:
Just want to jump in and say this is super neat! Seems like a nice and extensible way to build on top of the existing python syntax parser. I wish I had the time to be more involved right now but hopefully as my life gets less crazy in the next few months I'll have time to play around with this.

LeStarch:
@zimri-leisher yes, if `from future import annotations` will fix this in python pre 3.10, then yes!  If it doesn't, I can help you express these in earlier python syntax.

LeStarch:
Spelling is failing because of:

```
bytecode
expr
````

Both of these are valid, so they need to be added to `.github/actions/spelling/expect.txt`

File: pyproject.toml [hunk 1]
```diff
@@ -61,6 +61,7 @@ fprime-cli = "fprime_gds.executables.fprime_cli:main"
 fprime-seqgen = "fprime_gds.common.tools.seqgen:main"
 fprime-dp-write = "fprime_gds.executables.data_product_writer:main"
 fprime-gds = "fprime_gds.executables.run_deployment:main"
+fprime-seq = "fprime_gds.common.sequence.compiler:main"
```
LeStarch: Should we add `fpy` to this name to distinguish it? The old `fprime-seqgen` will lose to this in tab-complete and won't process correctly.

src/fprime_gds/common/sequence/README.md
```diff
@@ -0,0 +1,86 @@
+# FPy Advanced Sequencing Language Version 0.1
```
LeStarch: Eventually we should put this with F´.  We typically try to avoid directory-READMEs.

LeStarch (commented):
Python code looks nice.  I don't see any major issues.  Comments are a bit sparse, but otherwise this is nice!

src/fprime_gds/common/sequence/README.md
zimri-leisher: Happy to move it. Where/when should we move it?

pyproject.toml
zimri-leisher: Ah good point. Sure, if you think that's the best idea, we can make a generic "fpy" cmdline tool whose default behavior can be compiling.

zimri-leisher:
> Just want to jump in and say this is super neat! Seems like a nice and extensible way to build on top of the existing python syntax parser. I wish I had the time to be more involved right now but hopefully as my life gets less crazy in the next few months I'll have time to play around with this.

Thanks! Any feedback you have would be super valuable. Yes, I just realized that any language I was going to make was going to be heavily inspired by Python. This is just skipping the middle man of having to write some EBNF grammar etc etc.



> Python code looks nice. I don't see any major issues. Comments are a bit sparse, but otherwise this is nice!

The code is really the first version of this that works. I'll spend some time this week cleaning it up so I would feel comfortable merging it. Either way, most of this will have to be rewritten as the language evolves, but it was a helpful learning experience and proof-of-concept.

.github/actions/spelling/expect.txt
```diff
@@ -43,6 +47,12 @@ cdxoefg
 CFDG
 CFDP
 CFPD
+<<<<<<< HEAD
+cgi
+Chainable
+chans
+=======
+>>>>>>> 84bc482b1f8b088b1d6216401e46631363fc3c7b
```
thomas-bc: some git stuff snuck in here and below

pyproject.toml
```diff
@@ -61,6 +61,8 @@ fprime-cli = "fprime_gds.executables.fprime_cli:main"
...
+fpy-bytecode = "fprime_gds.common.fpy.serialize_bytecode:main"
```
thomas-bc: I think for a first pass we should keep an `fprime-` prefix, thoughts?

Or better (maybe for future PR), we could have subcommands like `fprime-fpy <asm|compile|etc.>` to have catalog of tools for Fpy. But I'm wary of GDS installing things in PATH that aren't prefixed with `fprime-*`.

src/fprime_gds/common/fpy/types.py
```diff
@@ -0,0 +1,86 @@
...
+    INVALID = 0
+    WAIT_REL = 0x00000001
+    WAIT_ABS = 0x00000002
+
+
+def time_type_from_json(js):
+    return TimeType(js["time_base"], js["time_context"], js["seconds"], js["useconds"])
+
+
+directives: list[StatementTemplate] = [
```
thomas-bc: Could we maybe rename this `FPY_DIRECTIVES` ? Since it's imported in another module

src/fprime_gds/common/fpy/types.py
```diff
@@ -0,0 +1,86 @@
...
+HEADER_FORMAT = "!BBBBBHI"
+HEADER_SIZE = struct.calcsize(HEADER_FORMAT)
+
+
+@dataclass
+class Header:
+    majorVersion: int
+    minorVersion: int
+    patchVersion: int
+    schemaVersion: int
```
thomas-bc, line 42: Just for my own curiosity, what is the intended usage of these version arguments down the road? Refuse to execute sequences that would not match FSW+Schema version ? Other things? I only see `schemaVersion` being used at the moment in FpySequencer

src/fprime_gds/common/fpy/types.py
```diff
@@ -0,0 +1,86 @@
...
+    DIRECTIVE = 0
+    CMD = 1
+
+
+@dataclass
+class StatementTemplate:
+    statement_type: StatementType
+    opcode: int
+    name: str
+    args: list[type[BaseType]]
```
thomas-bc, line 23: As you pointed out during the review, I think ValueType is what we're looking for here and in other uses of BaseType in this code. Represents a value and needs a constructor

src/fprime_gds/common/fpy/serialize_bytecode.py
```diff
@@ -0,0 +1,199 @@
...
+        return U32Type
+    elif type == "FwSizeStoreType":
+        return U16Type
+
+    raise RuntimeError("Unknown FPrime type alias " + str(type))
+
+
+def serialize_statement(stmt: StatementData) -> bytes:
+    # see https://github.com/nasa/fprime/issues/3023#issuecomment-2693051677
+    # TODO replace this with actual documentation
```
thomas-bc, line 40: The code is really easy to follow (thanks for that) so we don't need too much, but yes it'd be good to add a few docstrings to be able to understand what's going on without diving too deep.

thomas-bc (changes requested):
Looks great! Just a few nits and comments

One thing: when we were talking I was under the impression we were reviewing just the "bytecode" compilation, but I see both bytecode and `fpy` compilers in this PR. Do we intend to merge both at the same time? I've reviewed bytecode thoroughly and read through `fpy`. Let me know if you want me to do a review of `fpy`.

thomas-bc:
Also the CI is failing because of some type hint features that seem to have been introduced in Python3.10
```
    func: FpyType | FpyCmd | FpySeqDirective
   TypeError: unsupported operand type(s) for |: 'type' and 'type'
```
Since we intend to support 3.8+, we should use `typing.Union` for now

src/fprime_gds/common/fpy/types.py
zimri-leisher: Yes--I'm not super clear on them actually, but it was a design decision from @LeStarch and @timcanham for commonality with some heritage JPL systems.

pyproject.toml
thomas-bc: https://github.com/fprime-community/fprime-bootstrap/blob/devel/src/fprime_bootstrap/__main__.py shows a simple way of doing subcommands

zimri-leisher:
Question: if I add integration tests here that don't work on Ref, will that block merging?

thomas-bc (approved):

PR #184, opened 2024-12-19
Fixes nasa/fprime#3086. Default to insertion order for chrono hist

### LINKED ISSUE #3086: [Test API] Events With the Same Timestamp Get Miss-ordered by Chronological History
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

When ordering events in Chronological order with the same time stamp, the expectation is that the history falls back on received order. However, this is not the case.

LeStarch:
Chronological history needs to default to insertion order when the timestamps of the items are identical. This is fixed by changing the < to <= when comparing timestamps for "older" items.

| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes nasa/fprime#3086

### DIFF
--- a/src/fprime_gds/common/history/chrono.py
+++ b/src/fprime_gds/common/history/chrono.py
@@ -160,7 +160,9 @@ class ChronologicalHistory(History):
             the index that the item was inserted at (int)
         """
         for i, item in reversed(list(enumerate(ordered))):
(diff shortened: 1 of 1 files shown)


kevin-f-ortega (approved):
Good catch

PR #187, opened 2025-01-28
Add parameter json to .seq/.dat converter tool

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#3163  |

---
## Change Description

See nasa/fprime#3163

Attempt to use on Ref app.


LeStarch:
@zimri-leisher do you want to fix the spelling, or shall I?

zimri-leisher:
If there are no other comments, I will fix the spelling. However, I would prefer to do that after doing other code review things, so I only have to do it once.

LeStarch:
Agreed. We'll review and get back to you.

src/fprime_gds/common/tools/params.py
```diff
@@ -0,0 +1,233 @@
...
+        serialized += prm_instance_bytes
+    return serialized
+
+
+def parsed_json_to_seq(templates_and_values: list[tuple[PrmTemplate, dict]], include_save=False) -> list[str]:
+    cmds = []
+    cmds.append("; Autocoded sequence file from JSON")
+    for template_and_value in templates_and_values:
+        template, json_value = template_and_value
+        set_cmd_name = template.comp_name + "." + template.prm_name.upper() + "_PARAM_SET"
```
thomas-bc: I believe there was some issue with `_PARAM_SET` vs `_PRM_SET` (and SAVE) at some point. Is `PARAM` the correct name now? With `fprime@v3.6.0` it appears to be `PRM` in both XML and JSON dictionaries

src/fprime_gds/common/tools/params.py
```diff
@@ -0,0 +1,233 @@
...
+def parsed_json_to_dat(templates_and_values: list[tuple[PrmTemplate, dict]]) -> bytes:
+    serialized = bytes()
+    for template_and_value in templates_and_values:
+        template, json_value = template_and_value
+        prm_instance = instantiate_prm_type(json_value, template.prm_type_obj)
+
+        prm_instance_bytes = prm_instance.serialize()
+
+        # delimiter
+        serialized += b"\xA5"
```
thomas-bc, line 80: Could we add a docstring to that function and link for example to https://github.com/nasa/fprime/blob/devel/Svc/PrmDb/docs/sdd.md#32-functional-description (or any other resources you may have found useful) to understand what's happening here ?

src/fprime_gds/common/tools/params.py
```diff
@@ -0,0 +1,233 @@
...
+
+        record_size = 4 + len(prm_instance_bytes)
```
thomas-bc: I'm assuming 4 is the usual `FwPrmIdType` size here? It could be good to add a comment, or replace it with a VARIABLE. In the future this might even be configurable with a `--flag` passed to the CLI, since it's theoretically configurable by the project.

src/fprime_gds/common/tools/params.py
```diff
@@ -0,0 +1,233 @@
...
+def parse_json(json, name_dict: dict[str, PrmTemplate], include_implicit_defaults=False) -> list[tuple[PrmTemplate, dict]]:
+    """
+    json: the json object read from the .json file
+    name_dict: a dictionary of (fqn param name, PrmTemplate) pairs
+    include_implicit_defaults: whether or not to also include default values from the name dict
+                               if no value was specified in the json
+    @return a list of tuple of param template and the intended param value (in form of json dict)
+    """
+    # first, check the json for errors
+    for component_name in json:
```
thomas-bc: nit: could we rename this argument to e.g. `param_values_json` or anything that shows that is the user input and not the dictionary. `json` would also clash with the builtin `json` module if we were to not import it as `js` in the future somehow

thomas-bc (commented):
The code LGTM! It's awesome to see that you were able to build upon all the loaders/templates logic and all. 

Just a few minor nits below, and I also I have two questions:

1. Are you planning on documenting this tool? Regardless of the answer, could we add an example JSON input somewhere in the source code – that could help a lot both users and developers
2. Are you planning on adding some test cases?

Thanks!

src/fprime_gds/common/tools/params.py
zimri-leisher: Ah great, that's fixed now. I will update to PRM.

src/fprime_gds/common/tools/params.py
zimri-leisher: Sure, I can make a pass and add docstrings to everything

src/fprime_gds/common/tools/params.py
zimri-leisher: I will probably make it a const and add a comment. I don't think it's worth making an arg at the moment.

src/fprime_gds/common/tools/params.py
zimri-leisher: Yep good point, I can make a second pass on the function names too.

zimri-leisher:
@thomas-bc I responded to most of your feedback. Where should I add documentation for this tool?
Also, any thoughts on the cli name of this tool? Currently it's `fprime-params`

thomas-bc (commented):
LGTM! Do you want to fix spelling or shall I ?

To answer your questions:
- maybe name the CLI `fprime-paramdb-write` or something? We already have `fprime-dp-write` for data product binaries.
- Docs should go in https://github.com/nasa/fprime. Where is up to you... maybe as a How-To guide "How To initialize ParamDB" or something? If you just want to pop a quick markdown file in a PR I'll be happy to help put it in the correct location so that it displays nicely on the website and everything.

thomas-bc (approved):

PR #189, opened 2025-01-30
Upgrade spelling workflow 

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

v0.20 was using deprecated actions which GitHub now forbids to run. Upgrading to v0.24


thomas-bc:
Spelling workflow fails because it uses the old definition from the core repo. It does succeed when ran in my branch https://github.com/thomas-bc/fprime-gds/tree/spelling-workflow-upgrade

I recommend merging despite the failure here

LeStarch (approved):

PR #190, opened 2025-02-06
Fix internal JSON large number handling

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This fixes the big number (and other so called invalid) JSON object handling by pulling the `JSON.parse` and `JSON.stringify` implementations into a custom parser that intercepts the calls.

This parser replaces bad tokens with a custom object representation of the item, and then uses the `replacer` and `reviver` functions to restore the correct value.

## Rationale

`JSON.parse` cannot handle BigInt, NaN, Infinite, and null.  This side steps the issue by performing string manipulation to and from a looser specification, then using JSON, and then restoring the values from.

Try to break it with input from the box.

## Future Work

### DIFF
--- a/src/fprime_gds/flask/static/js/json.js
+++ b/src/fprime_gds/flask/static/js/json.js
@@ -28,7 +28,7 @@ class RegExLexer {
         // Null token
         ["NULL", /^null/],
         // NaN token
-        ["NULL", /^null/],
+        ["NAN", /^NaN/],
         // boolean token
         ["BOOLEAN", /^(true)|(false)/],
         // Open object token
@@ -148,7 +148,7 @@ export class SaferParser {
         "number",
         "bigint",
         null
-    ]
+    ];
 
 
     // Store the language variants the first time
@@ -288,7 +288,7 @@ export class SaferParser {
                     let replacement_object = {};
                     replacement_object[SaferParser.CONVERSION_KEY] = token_type;
                     replacement_object["value"] = token_text;
-                    return JSON.stringify(replacement_object)
+                    return SaferParser.language_stringify(replacement_object)
                 }
                 return token_text;
         });
--- a/src/fprime_gds/flask/static/js/json.js
+++ b/src/fprime_gds/flask/static/js/json.js
@@ -9,6 +9,78 @@
 
 
 /**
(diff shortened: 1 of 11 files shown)


thomas-bc:
Fixes nasa/fprime#3174

src/fprime_gds/flask/static/js/json.js
```diff
@@ -0,0 +1,275 @@
...
+    };
+    /**
+     * List of mapping tuples for clean parsing: string match, replacement type, and real (post parse) type
+     */
+    static MAPPINGS = [
+        [/(-Infinity)/, -Infinity],
+        [/(Infinity)/, Infinity],
+        [/(NaN)/, NaN],
+        [/(null)/, null],
+        [/( -?\d{10,})/, "bigint", convertInt]
```
thomas-bc: Need to make sure this regex is bullet proof

src/fprime_gds/flask/static/js/json.js
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     * Replace string occurrences of our gnarly types with a mapping equivalent
+     * @param string_value: value to replace
+     */
+    static replaceFromString(string_value) {
+        for (let i = 0; i < SaferParser.MAPPINGS.length; i++) {
+            let mapper = SaferParser.MAPPINGS[i];
+            string_value = string_value.replace(mapper[0], "{\"fprime{replacement\": \"$1\"}");
```
thomas-bc: Could use a variable instead of typing fprime{replacement

src/fprime_gds/flask/static/js/json.js
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     * Apply process function to raw json string only for data that is not qu
+     * @param json_string
+     * @param process_function
+     * @return {string}
+     */
+    static processUnquoted(json_string, process_function) {
+        // The initial state of any JSON string is unquoted
```
thomas-bc: Skip escaped quotes here

thomas-bc (approved):

PR #192, opened 2025-02-26
Add alias types to GDS

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**| [fprime/3272](https://github.com/nasa/fprime/issues/3272) |

---
## Change Description

Added alias types to GDS, updated Ref JSON dictionary and JSON loader tests to include alias types

## Rationale

Needed for https://github.com/nasa/fpp/issues/563

Added alias type test case to JSON loader tests, ran them successfully.

### DIFF
--- a/src/fprime_gds/common/loaders/json_loader.py
+++ b/src/fprime_gds/common/loaders/json_loader.py
@@ -102,6 +102,9 @@ class JsonLoader(dict_loader.DictLoader):
                 f"Dictionary type name has no corresponding type definition: {type_name}"
             )
 
+        if qualified_type.get("kind") == "alias":
+            return self.parse_type(qualified_type.get("underlyingType"))
+        
         if qualified_type.get("kind") == "array":
             return self.construct_array_type(type_name, qualified_type)
 
--- a/test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
+++ b/test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
@@ -1,6 +1,6 @@
 {
   "metadata" : {
-    "deploymentName" : "Ref",
+    "deploymentName" : "Ref.Ref",
     "projectVersion" : "TestVersion",
     "frameworkVersion" : "TestVersion",
     "libraryVersions" : [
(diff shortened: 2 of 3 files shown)


src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -102,6 +102,9 @@ def parse_type(self, type_dict: dict) -> BaseType:
                 f"Dictionary type name has no corresponding type definition: {type_name}"
             )
 
+        if qualified_type.get("kind") == "alias":
```
LeStarch, line 105: Please make sure I am following.  Once loaded into the dictionary, we are effectively saying "resolve the type" and all references within the loaded dictionary will ignore the alias to prefer the fully resolved type?

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
```diff
@@ -1,6 +1,6 @@
 {
   "metadata" : {
-    "deploymentName" : "Ref",
+    "deploymentName" : "Ref.Ref",
```
LeStarch, line 3: Shouldn't the deployment be just `Ref`?  Otherwise this is the *topology* name...

LeStarch (commented):
I have a few questions.

src/fprime_gds/common/loaders/json_loader.py
jwest115: Yes!

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
jwest115: The `deploymentName` field contains the fully-qualified topology name. In this case, the name of the topology is `Ref` and it exists in the `Ref` module.

src/fprime_gds/common/loaders/json_loader.py
LeStarch: This is the right approach for today.  However, it is good for me to understand where the aliasing is resolved because at some point users may ask for aliasing information to be reachable in the UI.  

To be clear, we are not ready for that plumbing to happen today and this approach is correct.

Thanks for clarifying.

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
LeStarch: This is a slight difference between how FPP views deployments and how F Prime/CMake views them.

That is FPP refers to the topology "as the deployment" where as F Prime refers to the build module because it does not know of model information in the build definitions.  This results in `Ref.Ref` vs `Ref`.

This isn't a problem, but I do appreciate the clarification.  It helps!

LeStarch (approved):

LeStarch:
@thomas-bc would you review too?

thomas-bc (approved):
LGTM!

PR #195, opened 2025-03-18
Updated api.py

ekswang:
| | |
|:---|:---|
|**_Related Issue(s)_**|  [3213](https://github.com/nasa/fprime/issues/3213)  |

---
## Change Description

A description of the changes contained in the PR.

An enhancement request for a Python tool that takes as input the `instances.fpp` file and creates the test/int directory within a deployment's directory and generate a Python configuration file in that directory. The configuration file will map components to their component instance name (this name is found in the `instances.fpp`); hence, `api.py` has been modified to take dictionary name of command, event, and channel (_instance.name_) without the need to pass in the deployment. In addition, an option to specify the Python configuration file has also been added in `pytest_integration.py`. Note, a Python script to that takes `instances.fpp` and creates a mapping (deployment/module/component) of all dictionary items has been created and ready to be checked into fprime repo if needed.

## Rationale

A rationale for this change. e.g. fixes bug, or most projects need XYZ feature.

Component instance names can be different across different deployments. As we start developing InT scripts for each component, we'll need to implement them while being deployment agnostic. The scripts will utilize the configuration file to know the component's instance name. This is different than the current approach to hardcode the instance name in the InT script.

- Created LedBlinker deployment
- Created a fprime-gds sandbox
- Ran LedBlinker system test with modifications to argument pass to api used on `fprime-fsw-0`

## Future Work

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -174,6 +174,7 @@ ipp
 itcl
 ixx
 janamian
+jawest
 Jdk
 jpl
 JSO
--- a/src/fprime_gds/common/gds_cli/base_commands.py
+++ b/src/fprime_gds/common/gds_cli/base_commands.py
@@ -130,7 +130,7 @@ class BaseCommand(abc.ABC):
             passes the given filter
         """
         project_dictionary = Dictionaries()
-        project_dictionary.load_dictionaries(dictionary_path, packet_spec=None)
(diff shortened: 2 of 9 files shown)


thomas-bc (commented):
I don't quite understand what the `component_config.py` is supposed to contain, and why we're adding it as a parser option? It doesn't look like you're using it right now? Is there code that the user is supposed to run, that this PR does not provide?

Let's discuss

PR #196, opened 2025-03-27
Add option to use JSON telemetry packet definitions

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**|  [fprime/3411](https://github.com/nasa/fprime/issues/3411) |

---
## Change Description

Adds CLI option `packet-set-name` for specifying the name of the packet set in the JSON dictionary to use when running the GDS with telemetry packets.

Tested with FPP telemetry packets and an updated Ref deployment that uses the TlmPacketizer.

### DIFF
--- a/src/fprime_gds/common/pipeline/dictionaries.py
+++ b/src/fprime_gds/common/pipeline/dictionaries.py
@@ -122,7 +122,7 @@ class Dictionaries:
             msg = f"[ERROR] Dictionary '{dictionary}' does not exist."
             raise Exception(msg)
         # Check for packet specification
-        if packet_set_name is not None:
+        if self._metadata["dictionary_type"] == "json" and packet_set_name is not None:
             packet_loader = fprime_gds.common.loaders.pkt_json_loader.PktJsonLoader(dictionary)
             self._packet_dict = packet_loader.get_id_dict(
                 None, packet_set_name, self._channel_name_dict
--- a/src/fprime_gds/common/pipeline/dictionaries.py
+++ b/src/fprime_gds/common/pipeline/dictionaries.py
@@ -14,6 +14,7 @@ from pathlib import Path
 import fprime_gds.common.loaders.ch_xml_loader
 import fprime_gds.common.loaders.cmd_xml_loader
 import fprime_gds.common.loaders.event_xml_loader
+import fprime_gds.common.loaders.pkt_json_loader
 import fprime_gds.common.loaders.pkt_xml_loader
 
 # JSON Loaders
@@ -48,7 +49,7 @@ class Dictionaries:
         self._versions = None
         self._metadata = None
 
-    def load_dictionaries(self, dictionary, packet_spec):
+    def load_dictionaries(self, dictionary, packet_spec, packet_set_name):
         """
         Loads the dictionaries based on the dictionary path supplied. Optional packet_spec is allowed to specify the
         definitions of packets.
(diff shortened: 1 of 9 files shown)


src/fprime_gds/common/pipeline/dictionaries.py
```diff
@@ -121,8 +122,13 @@ def load_dictionaries(self, dictionary, packet_spec):
             msg = f"[ERROR] Dictionary '{dictionary}' does not exist."
             raise Exception(msg)
         # Check for packet specification
-        if packet_spec is not None:
-            packet_loader = fprime_gds.common.loaders.pkt_xml_loader.PktXmlLoader()
+        if self._metadata["dictionary_type"] == "json" and packet_set_name is not None:
+            packet_loader = fprime_gds.common.loaders.pkt_json_loader.PktJsonLoader(dictionary)
```
thomas-bc, line 126: It is my expectation that most topologies (or at least in early development) will have no more than 1 packet set. In that case, it would be nice to pick that one and only packet set by default; and provide a meaningful error message if there are two or more packet sets. Thoughts?

src/fprime_gds/executables/cli.py
```diff
@@ -651,6 +651,14 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
                     "type": str,
                     "help": "Path to packet specification.",
                 },
+                ("--packet-set-name",): {
+                    "dest": "packet_set_name",
+                    "action": "store",
+                    "default": None,
+                    "required": False,
+                    "type": str,
+                    "help": "Packet set name",
```
thomas-bc: Can we add a few words explaining what this is and when it's needed? Shorter the better but this may confuse users who aren't familiar with it.

src/fprime_gds/executables/cli.py
```diff
@@ -651,6 +651,14 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
                     "type": str,
                     "help": "Path to packet specification.",
```
thomas-bc: We should call out on this one that this is now deprecated, preferring the JSON way.

thomas-bc (changes requested):

src/fprime_gds/common/pipeline/dictionaries.py
thomas-bc: Wait I think I may have misunderstood what packet sets are when reviewing this
https://nasa.github.io/fpp/fpp-spec.html#Specifiers_Telemetry-Packet-Set-Specifiers

Should we not import **all** packet sets instead?

jwest115: Topologies can have multiple packet sets. I am thinking that for now, we keep the `packet-set-name` option and _do not_ include a default if there is one packet set in the dictionary since that is most clear and matches what we currently support in the XML. For the future, I think we should add an ID to the packet set name in FPP and downlink the packet set ID and packet ID so the GDS can decode packets from any packet set.

src/fprime_gds/executables/cli.py
jwest115: Sure, I'll update this.

src/fprime_gds/executables/cli.py
jwest115: Will do!

src/fprime_gds/common/pipeline/dictionaries.py
thomas-bc: Sounds good, thanks for clarifying.

thomas-bc (approved):
LGTM - thanks!

PR #197, opened 2025-04-09
F Prime CI/CD Test Runner

ekswang:
| | |
|:---|:---|
|**_Related Issue(s)_**| [2646](https://github.com/nasa/fprime/issues/2646) |
|**_Documentation Included (y/n)_**| N/A |

---
## Change Description

Test Runner has been updated to collect the following files:
    - Copies of FSW dictionaries
    - A copy of FSW binary
    - Test report

## Rationale

This is a feature request to include specified files in the output directory of the test runner.

N/A

## Future Work

N/A

### DIFF
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -69,8 +69,10 @@ class IntegrationTestAPI(DataHandler):
         if logpath is not None:
             base_dir = Path(self.pipeline.dictionary_path).parents[1]
             for subdir in ['bin', 'dict']:
-                shutil.copytree(base_dir / subdir, Path(logpath) / subdir,
-                                dirs_exist_ok=True)
+                dir_path = base_dir / subdir
+                if dir_path.is_dir():
+                    shutil.copytree(dir_path, Path(logpath) / subdir,
+                                    dirs_exist_ok=True)
 
         # A predicate used as a filter to choose which events to log automatically
         self.event_log_filter = self.get_event_pred()
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -9,7 +9,6 @@ telemetry and dictionaries.
 """
 import signal
 import time
-import os
 from pathlib import Path
 import shutil
 
@@ -69,8 +68,7 @@ class IntegrationTestAPI(DataHandler):
         # Copy dictionaries and binary file to output directory
         base_dir = Path(self.pipeline.dictionary_path).parents[1]
         for subdir in ['bin', 'dict']:
-            shutil.copytree(os.path.join(str(base_dir), subdir),
-                            os.path.join(logpath, subdir),
+            shutil.copytree(base_dir / subdir, Path(logpath) / subdir,
                             dirs_exist_ok=True)
 
         # A predicate used as a filter to choose which events to log automatically
@@ -231,7 +229,7 @@ class IntegrationTestAPI(DataHandler):
         Returns:
             The name of the deployment (str)
         """
-        return os.path.basename(str(Path(self.pipeline.dictionary_path).parents[1]))
+        return Path(self.pipeline.dictionary_path).parent.parent.name
 
     ######################################################################################
     #   History Functions
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -66,10 +66,11 @@ class IntegrationTestAPI(DataHandler):
         self.logger = TestLogger(logpath) if logpath is not None else None
 
         # Copy dictionaries and binary file to output directory
-        base_dir = Path(self.pipeline.dictionary_path).parents[1]
-        for subdir in ['bin', 'dict']:
-            shutil.copytree(base_dir / subdir, Path(logpath) / subdir,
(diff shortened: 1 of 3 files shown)


src/fprime_gds/common/testing_fw/api.py
```diff
@@ -215,6 +225,14 @@ def set_event_log_filter(
         """
         self.event_log_filter = self.get_event_pred(event, args, severity, time_pred)
 
+    def get_deployment(self):
+        """
+        Get the deployment of the target using the loaded FSW dictionary path
+        Returns:
+            The name of the deployment (str)
+        """
+        return os.path.basename(str(Path(self.pipeline.dictionary_path).parents[1]))
```
thomas-bc: The Path object should also have a `.name` attribute which is equivalent

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -63,6 +66,13 @@ def __init__(self, pipeline, logpath=None, fsw_order=True):
         # Initialize the logger
         self.logger = TestLogger(logpath) if logpath is not None else None
 
+        # Copy dictionaries and binary file to output directory
+        base_dir = Path(self.pipeline.dictionary_path).parents[1]
+        for subdir in ['bin', 'dict']:
+            shutil.copytree(os.path.join(str(base_dir), subdir),
```
thomas-bc: We tend to use the Path `/` operator rather than `os.path.join`, could we use that here and below?

`basedir / subdir` should work?

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -9,6 +9,9 @@
 """
 import signal
 import time
+import os
```
thomas-bc: Can remove `import os` if we replace all the calls by the pathlib equivalents

src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -38,7 +39,19 @@ def pytest_addoption(parser):
...
+    # Add an option to specify JUnit XML report file
+    parser.addoption(
+        "--junit-xml-file",
+        action="store",
+        default="report.xml",
+        help="File to store JUnit XML report. [default: %(default)s]",
+    )
+
+def pytest_configure(config):
+    """ Create a JUnit XML report file to capture the test result """
```
thomas-bc: Could we start this docstring by saying this is a pytest hook ?
https://docs.pytest.org/en/stable/reference/reference.html#pytest.hookspec.pytest_configure

Other devs might benefit from knowing they can add stuff there if needed

thomas-bc (commented):
This looks good! Just a few pedantic styling nitpicks... Thanks!

thomas-bc:
@kevin-f-ortega would you please review and confirm this is the behavior you expected?

src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -38,7 +38,24 @@ def pytest_addoption(parser):
...
+
+def pytest_configure(config):
+    """ This is a hook to allow plugins and conftest files to perform initial configuration
+    
+    This hook is called for every initial conftest file after command line options have been parsed. After that, the 
+    hook is called for other conftest files as they are registered.
+    """
+    # Create a JUnit XML report file to capture the test result
+    config.option.xmlpath = Path(config.getoption("--logs")) / config.getoption("--junit-xml-file")
+    config.option.junitxml = True
```
thomas-bc, line 58: Could you please link to some documentation on what these specific options are? I can't seem to find anything

Or are they just values you are using yourself in some downstream script ?

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -63,6 +65,12 @@ def __init__(self, pipeline, logpath=None, fsw_order=True):
         # Initialize the logger
         self.logger = TestLogger(logpath) if logpath is not None else None
 
+        # Copy dictionaries and binary file to output directory
+        base_dir = Path(self.pipeline.dictionary_path).parents[1]
+        for subdir in ['bin', 'dict']:
+            shutil.copytree(base_dir / subdir, Path(logpath) / subdir,
```
kevin-f-ortega: You should check if `logpath` is `None`. I don't think we want to copy over if logpath is None. What do you think?

ekswang: Ok. I've added a check for that.

kevin-f-ortega (dismissed):
LGTM

src/fprime_gds/common/testing_fw/pytest_integration.py
ekswang: By default output JUnit xml file is not turned on, so these values are used to enable the output and redirect the output and save it under a specified file name.
https://docs.pytest.org/en/stable/how-to/output.html
https://docs.pytest.org/en/stable/_modules/_pytest/junitxml.html
https://github.com/pytest-dev/pytest/blob/2439729413e701a19398316711ec5d795ae2ec63/src/_pytest/junitxml.py#L379-L390

thomas-bc (approved):

kevin-f-ortega (dismissed):

thomas-bc (changes requested):
It looks like tests are failing – would you look into it?

thomas-bc (approved):

PR #198, opened 2025-04-23
Add Python 3.12 and 3.13 to CI runs

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Adding Python 3.12 and 3.13 to CI runs

### DIFF
--- a/src/fprime_gds/executables/fprime_cli.py
+++ b/src/fprime_gds/executables/fprime_cli.py
@@ -14,7 +14,7 @@ from copy import deepcopy
 from typing import Callable, List, Union
 
 import argcomplete
-import pkg_resources
+import importlib.metadata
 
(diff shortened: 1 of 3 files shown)


src/fprime_gds/executables/fprime_cli.py
```diff
@@ -14,7 +14,7 @@
 from typing import Callable, List, Union
 
 import argcomplete
-import pkg_resources
```
thomas-bc, line 17: `pkg_resources` was not available in Python3.12 and causing problems in CI. Replaced with `importlib` which is available in the std library for Python 3.8+ and is the recommended way of getting package versions

LeStarch (approved):

PR #199, opened 2025-04-25
Use Fw types defined in JSON dictionary

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**| [#3406](https://github.com/nasa/fprime/issues/3406) |
|**_Has Unit Tests (y/n)_**| Y |
|**_Documentation Included (y/n)_**| N |

---
## Change Description

Updated GDS to use the Fw types defined in the JSON dictionary (if they exist), otherwise use the defaults defined in ConfigManager.

Added test case for checking the JSON decoding aspect. Tested the GDS with the Ref deployment and confirmed Fw types defined in the dictionary were used by the GDS.


src/fprime_gds/common/loaders/fw_type_json_loader.py
```diff
@@ -0,0 +1,61 @@
...
+
+    TYPE_DEFINITIONS_FIELD = "typeDefinitions"
+    
+    # Dictionary mapping from JSON dictionary type qualified name to GDS key in config dictionary
+    FW_TYPE_NAME_MAPPING = {
+        "FwPacketDescriptorType": "msg_desc",
+        "FwChanIdType": "ch_id",
+        "FwEventIdType": "event_id",
+        "FwOpcodeType": "op_code",
+        "FwTlmPacketizeIdType": "pkt_id"
```
thomas-bc: Do you think it would help / be possible to rename to GDS keys to match the typenames in the JSON dictionary ?

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
```diff
@@ -8,6 +8,86 @@
     "dictionarySpecVersion" : "1.0.0"
   },
   "typeDefinitions" : [
+    {
+      "kind" : "alias",
+      "qualifiedName" : "FwTlmPacketizeIdType",
```
thomas-bc, line 13: How come only these types are in the JSON dictionary, when it seems there's a lot more in `FpConfig.fpp` ?

src/fprime_gds/common/loaders/fw_type_json_loader.py
thomas-bc: This ties back to the question about why so few elements in the JSON dict - can this go out of date if we decide we need to ingest `FwDpIdType` into the config or something?

src/fprime_gds/common/pipeline/dictionaries.py
```diff
@@ -76,6 +79,15 @@ def load_dictionaries(self, dictionary, packet_spec, packet_set_name):
...
             self._channel_name_dict = json_channel_loader.get_name_dict(None)
             self._channel_id_dict = json_channel_loader.get_id_dict(None)
+            # Fw Types
+            fw_types_loader = fprime_gds.common.loaders.fw_type_json_loader.FwTypeJsonLoader(
+                dictionary
+            )
+            self._fw_type_name_dict = fw_types_loader.get_name_dict(None)
+            if config:
+                # Update config to use Fw types defined in the JSON dictionary
+                for fw_type_name, fw_type in self._fw_type_name_dict.items():
```
thomas-bc: Thoughts on refactoring this a little so that config doesn't get modified here, but rather for example in StandardPipeline where it is needed?

src/fprime_gds/common/pipeline/standard.py
```diff
@@ -84,7 +84,7 @@ def setup(
         self.distributor = fprime_gds.common.distributor.distributor.Distributor(config)
         self.client_socket = self.__transport_type()
         # Setup dictionaries encoders and decoders
-        self.dictionaries.load_dictionaries(self.dictionary_path, packet_spec, packet_set_name)
+        self.dictionaries.load_dictionaries(self.dictionary_path, packet_spec, packet_set_name, config)
         self.coders.setup_coders(
```
thomas-bc, line 102: Could do something like
```
# pseudocode
config.update(self.dictionaries.fw_types)
```
instead of modifying config in in `load_dictionaries`.

thomas-bc (changes requested):

thomas-bc (changes requested):
Just a few questions/comments

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
jwest115: For the JSON dictionary, I only require the types that are needed by the GDS. Currently, the GDS only uses (and allows the user to override) the following types: https://github.com/nasa/fprime-gds/blob/devel/src/fprime_gds/common/utils/config_manager.py#L154-L162

src/fprime_gds/common/pipeline/dictionaries.py
jwest115: Agreed!

src/fprime_gds/common/pipeline/standard.py
jwest115: Agreed! Will make that change.

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
jwest115: We could expand the types that are required by the dictionary, but they would not be used in the GDS (yet?).

src/fprime_gds/common/loaders/fw_type_json_loader.py
jwest115: Good suggestion! I will make the GDS keys match the type names from the JSON dictionary. Added my response to your second question below.

src/fprime_gds/common/loaders/fw_type_json_loader.py
```diff
@@ -0,0 +1,61 @@
...
+        name_dict = {}
+
+        if self.TYPE_DEFINITIONS_FIELD not in self.json_dict:
+            raise GdsDictionaryParsingException(
+                f"Ground Dictionary missing '{self.TYPE_DEFINITIONS_FIELD}' field: {str(self.json_file)}"
+            )
+
+        for type_def in self.json_dict[self.TYPE_DEFINITIONS_FIELD]:
+            try:
+                if type_def["kind"] == "alias" and type_def["qualifiedName"] in self.FW_TYPE_NAMES:
```
thomas-bc: Would it make sense to match on whether the name starts with `Fw`, rather than maintaining a list?

thomas-bc: And maybe it being unqualified (contains no `.`)  ?

src/fprime_gds/common/pipeline/standard.py
```diff
@@ -93,7 +94,11 @@ def setup(
         # Setup dictionaries encoders and decoders
         self.dictionaries.load_dictionaries(
             self.dictionary_path, packet_spec, packet_set_name
-        )
+        )        
+        # Update config to use Fw types defined in the JSON dictionary
+        if self.dictionaries._fw_type_name_dict:
+            for fw_type_name, fw_type in self.dictionaries._fw_type_name_dict.items():
```
thomas-bc: We should add and access a `@property` in the `dictionaries` object rather than accessing the `_fw_type_name_dict`.

https://github.com/nasa/fprime-gds/blob/ba05c6d2059506cfd248292026010136aff15c1b/src/fprime_gds/common/pipeline/dictionaries.py#L138-L166

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
thomas-bc: @jwest115 I see, I guess this is just a test dictionary – but I would expect the actual dictionary generated by `fprime/Ref/` to include all type aliases, and not only the ones needed for the GDS, is that correct?

src/fprime_gds/common/loaders/fw_type_json_loader.py
thomas-bc: The GDS doesn't need them for now, but we could load them nonetheless. That way no need to maintain the list, and who knows maybe a user developing a GDS extension might need the information for some reason.

thomas-bc (changes requested):
A few nits, let me know!

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
jwest115: Nope! FPP only requires the the following types when building the dictionary:
- FwChanIdType
- FwEventIdType 
- FwOpcodeType
- FwPacketDescriptorType
- FwTlmPacketizeIdType

Reason we went with this set is because they are the types that are configurable in FPrime and the GDS (and the GDS needs to know what they are in order to decode things, etc.).

If the user defines other types and uses them in commands, events, etc. then yes, they will be included in the dictionary. The set above is just the types we _require_.

LeStarch: Do you mean to say that anything used in Commands/Events/Telemetry will be included because they are used as part of other dictionary objects and that the above set is the core set of types that describe the packets themselves and aren't strictly contained within the model of an element (e.g. command) within those packets?

What if the user defines another packet protocol (e.g. CCSDS) and needs to have FPP constants that influence Flight and Ground but that does not end up a modeled type (e.g. command)?

LeStarch: Apart from the extra amount of data stored in the dictionary, why wouldn't we include all non-opaque types (modeled types) in the dictionary?  If we know the type's definition, why not provide it?

jwest115: > Do you mean to say that anything used in Commands/Events/Telemetry will be included because they are used as part of other dictionary objects and that the above set is the core set of types that describe the packets themselves and aren't strictly contained within the model of an element (e.g. command) within those packets?

Yes, that sounds right!

> What if the user defines another packet protocol (e.g. CCSDS) and needs to have FPP constants that influence Flight and Ground but that does not end up a modeled type (e.g. command)?

In this implementation, for the example you provided (a user defines another packet protocol and defines some FPP constant), the type would _not_ be included in the dictionary since it does not appear in some command/event/etc.

src/fprime_gds/common/loaders/fw_type_json_loader.py
jwest115: Done!

test/fprime_gds/common/loaders/resources/RefTopologyDictionary.json
jwest115: The current dictionary tool does not dump all type information to the JSON dictionary, it only outputs the type information if the type is used by a dictionary object (command/event/etc), but I suppose it can. Do you have any thoughts on doing (or maybe not doing) this @bocchino?

thomas-bc (approved):
LGTM! Need to coordinate with https://github.com/nasa/fpp/pull/686 for releasing in sync

PR #200, opened 2025-04-26
Plugin Updates

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

1. Added `DataHandler` plugins to handle most common use case of "I want to intercept and do X with F Prime data".  
2. Added helper to allow for start-up apps to get the standard pipeline easily
3. Added `@gds_plugin(PluginClass)` decorator for ease in defining plugins
4. Refactored Plugin CLI parsing

Fixed bugs:
1. Plugin parsing no longer reinstantiates plugins multiple times
2. Plugin instantiation is left to the user
3. Plugin parsing no long requires **all plugins** in systems that are running with less plugins (e.g. comms uses 2: communication and framing)

### DIFF
--- a/src/fprime_gds/common/handlers.py
+++ b/src/fprime_gds/common/handlers.py
@@ -6,6 +6,7 @@ defines the "DataHandler" base class for handling data.
 
 @author mstarch
 """
+
 import abc
 from typing import List, Type
 from fprime_gds.plugin.definitions import gds_plugin_specification, gds_plugin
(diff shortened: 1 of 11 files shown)


src/fprime_gds/common/handlers.py
```diff
@@ -27,6 +30,42 @@ def data_callback(self, data, sender=None):
         """
 
 
+class DataHandlerPlugin(DataHandler, abc.ABC):
+    """PLugin class allowing for custom data handlers
```
thomas-bc, line 34: (Just mentioning since we talked about it online: ) I actually think it's probably better not to duplicate the `@abstractmethod`, like you currently have it – users who read through the code should be able to figure out how to follow the class tree. We should make sure to document in bright bold font that to implement a plugin, one needs to make sure to implement abstract methods of all parents. And I'm sure it errors out if you don't ?

thomas-bc (approved):
This all looks good to me! Do you intend to document these new plugins as part of ROSES ?

src/fprime_gds/common/handlers.py
LeStarch: Yes, I have a follow-up PR to document this as part of ROSES.  I will evaluate then if this can be fixed by documentation, moving/cloning the method, or a bit of both.

PR #202, opened 2025-05-01
Add await_connection to Integration Test API

ekswang:
Added wait_for_dataflow()

| | |
|:---|:---|
|**_Related Issue(s)_**|  [2642](https://github.com/nasa/fprime/issues/2642) |

---
## Change Description

There is a need to add a function in the fprime_test_api that allows tests to block or wait on data flow.

## Rationale

This function allows user to launch fit test calling fprime_test_api.wait_for_dataflow() and be able to bring up the flight software within a specified time duration.

- Launch fprime-gds without the binary being loaded via one terminal
- Load the binary via a separate terminal and verify GSE is up and running
- Execute fit test
- While fit test is executing, kill the process that's running the binary
- Verify wait_for_dataflow() asserts when detecting no data flow for specified time duration
- Load the binary via a separate terminal and verify GSE is up and running
- Verify wait_for_dataflow() continues after detecting data flow

## Future Work

N/A

### DIFF
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -234,6 +234,31 @@ class IntegrationTestAPI(DataHandler):
         """
         return Path(self.pipeline.dictionary_path).parent.parent.name
 
+    def wait_for_dataflow(self, count=1, start=None, timeout=120):
+        """
+        Wait for data flow by checking for any telemetry updates within a specified timeout.
+
+        Args:
+            count: either an exact amount (int) or a predicate to specify how many objects to find
+            start: an optional index or predicate to specify the earliest item to search
+            timeout: the number of seconds to wait before terminating the search (int)
+        """
+        if start is None:
+            start = self.get_latest_time()
+
+        chan_dict = self.pipeline.dictionaries.channel_name
+        chans = list(chan_dict.keys())
+
+        chan_hist = self.get_telemetry_subhistory()
+        result = self.await_telemetry_count(
+            count, channels=chans, history=chan_hist, start=start, timeout=timeout
+        )
+        if not result:
+            msg = f'Failed to detect any data flow for {timeout} s.'
+            self.__log(msg, TestLogger.RED)
+            assert False, msg
+        self.remove_telemetry_subhistory(chan_hist)
+
     ######################################################################################
     #   History Functions
     ######################################################################################
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -234,20 +234,18 @@ class IntegrationTestAPI(DataHandler):
         """
         return Path(self.pipeline.dictionary_path).parent.parent.name
 
-    def wait_for_dataflow(self, count=1, start=None, timeout=120):
+    def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
         """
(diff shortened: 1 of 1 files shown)


src/fprime_gds/common/testing_fw/api.py
```diff
@@ -234,6 +234,31 @@ def get_deployment(self):
...
+        """
+        if start is None:
+            start = self.get_latest_time()
+
+        ch_dict = self.pipeline.dictionaries.channel_name
+        channels = list(ch_dict.keys())
+
+        history = self.get_telemetry_subhistory()
+        result = self.await_telemetry_count(
+            count, channels=channels, history=history, start=start, timeout=timeout
```
thomas-bc, line 252: Does leaving `channels=None` accomplish the same result?

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -234,6 +234,31 @@ def get_deployment(self):
...
+            start: an optional index or predicate to specify the earliest item to search
+            timeout: the number of seconds to wait before terminating the search (int)
+        """
+        if start is None:
+            start = self.get_latest_time()
+
+        ch_dict = self.pipeline.dictionaries.channel_name
+        channels = list(ch_dict.keys())
+
+        history = self.get_telemetry_subhistory()
```
thomas-bc, line 250: Why create a subhistory and not use the `get_telemetry_test_history` ?

ekswang: Creating a subhistory will make sure that there's no residual/past (last instance/packet) data being returned due to the usage of getting the latest time, which gets the time from the last telemetry data,. Hence, await_telemetry_count function would have returned that object/data even when there's no data flow.

src/fprime_gds/common/testing_fw/api.py
ekswang: Yes, you are right. I just looked at ind_history_count() and it will default to search all if None is passed into the function. I will remove the unnecessary definition of channels.

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -234,6 +234,29 @@ def get_deployment(self):
...
+            channels: a channel specifier or list of channel specifiers (mnemonic, ID, or predicate). All will count if None
+            start: an optional index or predicate to specify the earliest item to search
+            timeout: the number of seconds to wait before terminating the search (int)
+        """
+        if start is None:
+            start = self.get_latest_time()
+
+        history = self.get_telemetry_subhistory()
+        result = self.await_telemetry_count(
+            count, channels=channels, history=history, start=start, timeout=timeout
```
thomas-bc, line 252: I think you missed changing to `channels=None` on your latest commit here

thomas-bc (changes requested):

src/fprime_gds/common/testing_fw/api.py
thomas-bc: Sounds good - thanks for the explanation!

src/fprime_gds/common/testing_fw/api.py
ekswang: I set it to channels=None as default in the function. Just to be flexible for user.

thomas-bc: Ohhh sorry my bad - thanks!

thomas-bc (approved):

PR #203, opened 2025-05-07
Disable the generation of junitxml report by default

ekswang:
| | |
|:---|:---|
|**_Related Issue(s)_**| [2646](https://github.com/nasa/fprime/issues/2646) |

---
## Change Description

This is a hotfix to an existing feature request documented in ticket #197. Previous change updated the default pytest behavior. This hotfix will restore the original pytest behavior.

## Rationale

Running pytest should not generate junitxml report. The generation of this report should only be requested or set by user.

- Verified the generation of the junitxml report is done only when the `--gen-junitxml` flag is set or when the file location is passed in `--junit-xml` 

## Future Work

N/A


LeStarch (approved):
I agree this will do what is said.  Thanks!

PR #205, opened 2025-05-14
FpySequencer bytecode serializer V0.2

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#3521 |
|**_Documentation Included (y/n)_**|  y|

---
## Change Description

Adds all bytecode directives necessary for compatibility with the latest FpySequencer.

### DIFF
--- a/src/fprime_gds/common/fpy/serialize_bytecode.py
+++ b/src/fprime_gds/common/fpy/serialize_bytecode.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from dataclasses import astuple
+import inspect
 import json
 from pathlib import Path
 from argparse import ArgumentParser
@@ -13,25 +14,17 @@ from fprime_gds.common.fpy.types import (
     HEADER_FORMAT,
     FOOTER_FORMAT,
     StatementType,
-    FPY_DIRECTIVES
+    FPY_DIRECTIVES,
+    BytecodeParseContext,
+    get_type_obj_for,
 )
+from fprime_gds.common.loaders.ch_json_loader import ChJsonLoader
 from fprime_gds.common.loaders.cmd_json_loader import CmdJsonLoader
 from fprime.common.models.serialize.numerical_types import (
     U8Type,
-    U16Type,
-    U32Type,
 )
-from fprime.common.models.serialize.type_base import ValueType
 
-
-def get_type_obj_for(type: str) -> type[ValueType]:
-    """returns a type object representing the ValueType that corresponds to a type alias"""
-    if type == "FwOpcodeType":
-        return U32Type
-    elif type == "FwSizeStoreType":
-        return U16Type
-
-    raise RuntimeError("Unknown FPrime type alias " + str(type))
+from fprime_gds.common.loaders.prm_json_loader import PrmJsonLoader
 
 
 def serialize_statement(stmt: StatementData) -> bytes:
@@ -59,7 +52,7 @@ def serialize_statement(stmt: StatementData) -> bytes:
 
 
 def parse_str_as_statement(
-    stmt: str, templates: list[StatementTemplate]
+    stmt: str, templates: list[StatementTemplate], context: BytecodeParseContext
 ) -> StatementData:
     """Converts a human-readable line of bytecode into a StatementData instance, given a list of
     possible statement templates"""
@@ -95,7 +88,12 @@ def parse_str_as_statement(
         )
     for index, arg_json in enumerate(args):
         arg_type = matching_template.args[index]
-        arg_value = arg_type(arg_json)
+        if inspect.isclass(arg_type):
+            # it's a type. instantiate it with the json
+            arg_value = arg_type(arg_json)
+        else:
+            # it's a function. give it the json and the ctx
+            arg_value = arg_type(arg_json, context)
         arg_values.append(arg_value)
 
     return StatementData(matching_template, arg_values)
@@ -112,7 +110,7 @@ def main():
         "--dictionary",
         type=Path,
         help="The JSON topology dictionary to compile against",
-        required=True
+        required=True,
     )
 
     arg_parser.add_argument(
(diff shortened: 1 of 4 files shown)


src/fprime_gds/common/fpy/types.py
```diff
@@ -1,11 +1,28 @@
...
+from fprime.common.models.serialize.numerical_types import U32Type, U16Type, U8Type
+from fprime.common.models.serialize.string_type import StringType
+
+from fprime_gds.common.loaders.json_loader import PRIMITIVE_TYPE_MAP
+from fprime_gds.common.templates.ch_template import ChTemplate
+from fprime_gds.common.templates.prm_template import PrmTemplate
+
+
+def get_type_obj_for(type: str) -> type[ValueType]:
+    if type == "FwOpcodeType":
```
LeStarch, line 17: This could be rewritten to use type alias information in the dictionary.  Not a blocker.

LeStarch (dismissed):

src/fprime_gds/common/fpy/types.py
zimri-leisher: Happy to do this if you make a ticket and assign it to me

src/fprime_gds/common/fpy/serialize_bytecode.py
```diff
@@ -155,27 +154,65 @@ def serialize_bytecode(input: Path, dictionary: Path, output: Path=None):
         )
         stmt_templates.append(stmt_template)
 
-    stmts = []
+    tlm_json_loader = ChJsonLoader(str(dictionary))
+    (cmd_id_dict, cmd_name_dict, versions) = tlm_json_loader.construct_dicts(
```
thomas-bc: Can we rename this to `ch_id_dict`? It confused me for a second with 
```
context.channels = cmd_name_dict
```
below

Probably copy/paste remnants. Also could be good to use `_` name for unused values if you think that helps with readability

src/fprime_gds/common/fpy/serialize_bytecode.py
```diff
@@ -155,27 +154,65 @@ def serialize_bytecode(input: Path, dictionary: Path, output: Path=None):
...
+    context = BytecodeParseContext()
+    context.types = cmd_json_dict_loader.parsed_types
+    context.channels = cmd_name_dict
+    context.params = prm_name_dict
+
+    input_lines = input.read_text().splitlines()
+    input_lines = [line.strip() for line in input_lines]
+    # remove comments and empty lines
+    input_lines = [
+        line for line in input_lines if not line.startswith(";") and len(line) > 0
```
thomas-bc, line 176: Just confirming - are "until end-of-line" comments not supported?
```
; this is a supported comment
<valid statement> ; this is not supported ?
```

src/fprime_gds/common/fpy/serialize_bytecode.py
```diff
@@ -155,27 +154,65 @@ def serialize_bytecode(input: Path, dictionary: Path, output: Path=None):
...
+    ]
+
+    goto_tags = {}
+    statement_idx = 0
+    statement_strs: list[str] = []
+    for stmt_idx, stmt in enumerate(input_lines):
```
thomas-bc: Should this be renamed `line_idx` or `_` (appears unused) ? Otherwise there's `statement_idx` and `stmt_idx` which is confusing

src/fprime_gds/common/fpy/types.py
```diff
@@ -78,4 +169,34 @@ def time_type_from_json(js):
         "WAIT_ABS",
         [time_type_from_json],
     ),
+    StatementTemplate(
+        StatementType.DIRECTIVE,
+        DirectiveOpcode.SET_LVAR.value,
+        "SET_LVAR",
+        [U8Type, arbitrary_type_from_json],
```
thomas-bc, line 177: Bit of a nit, but should we change `StatementTemplate`'s type hint to be `args: list[(type[BaseType] | function-that-returns-a-BaseType-instance)]` (not sure if there's a syntax for that actually) ? Or call it out in the docstring

src/fprime_gds/common/fpy/types.py
```diff
@@ -55,16 +73,89 @@ class Footer:
...
+    return U32Type(js)
+
+
+def tlm_chan_id_from_json(js, ctx: BytecodeParseContext):
+    if isinstance(js, str):
+        if js not in ctx.channels:
+            raise RuntimeError("Unknown telemetry channel " + str(js))
+        return get_type_obj_for("FwChanIdType")(ctx.channels[js].id)
+    elif isinstance(js, int):
+        matching = [tmp for tmp in ctx.channels.keys() if tmp.id == js]
```
thomas-bc, line 136: There's two Python dicts provided by the JsonDictLoaders. One indexed by channel name,  another one indexed by channel ID. Is there a reason you're walking through the name_dict to find the matching ID, rather than just doing a lookup in the id_dict ? If not, it's not a blocker, just to keep in mind for the future.

thomas-bc (changes requested):
A couple questions and nits - I hope you don't mind

src/fprime_gds/common/fpy/serialize_bytecode.py
zimri-leisher: yeah... i was a little too lazy to go parse all the parentheses and strings from the json. is there an easy way to do it?

thomas-bc: Not that I know of, and would probably be error-prone - this is very good, I just wanted to confirm

src/fprime_gds/common/fpy/types.py
zimri-leisher: eh i think it would just be slightly more confusing to include both the name and id dict in the context... it's not that hard to find by id

zimri-leisher:
Feedback responded to @thomas-bc

src/fprime_gds/common/fpy/types.py
thomas-bc: You should be able to retrieve the types from the config object, like shown in this PR https://github.com/nasa/fprime-gds/pull/199

zimri-leisher: Ah, I can make this change now if you'd like. Was initially thinking could come in 0.3 because I don't want to block the next FPrime release.

thomas-bc: Yes no worries, let's table that to v0.3

thomas-bc (approved):

thomas-bc:
Thanks Zimri!

PR #208, opened 2025-05-30
Adding skip-check argument to uart adapter

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Allows users of the uart adapter to turn-off the check for existence.  This is helpful when needing to open comms before the UART device is booted.


thomas-bc (approved):
LGTM!

PR #210, opened 2025-06-05
Adding config driven parser and placing it on UI

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Add config driven parser, and config for GDS!

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -209,7 +209,7 @@ class ParserBase(ABC):
             arguments: arguments to process, None to use command line input
         Returns: namespace with all parsed arguments from all provided ParserBase subclasses
         """
-        return cls._parse_args(lambda parse: parse.parse_known_args, parser_classes, description, arguments, **kwargs)
+        return cls._parse_args(parser_classes, description, arguments, allow_unkowns=True, **kwargs)
 
     @classmethod
     def parse_args(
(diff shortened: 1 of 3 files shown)


src/fprime_gds/executables/cli.py
```diff
@@ -19,6 +19,9 @@
 import platform
 import re
 import sys
+
+import yaml
```
thomas-bc, line 23: Should probably add `PyYAML` as an explicit dependency in https://github.com/nasa/fprime-gds/blob/devel/pyproject.toml

src/fprime_gds/executables/cli.py
```diff
@@ -185,29 +188,85 @@ def handle_arguments(self, args, **kwargs):
...
+        arguments methods will be called, and the final namespace will be returned. This will allow unknown arguments
+        which are returned as the last tuple result.
+
+        Args:
+            parser_classes: a list of ParserBase subclasses that will be used to
+            description: description passed ot the argument parser
+            arguments: arguments to process, None to use command line input
+        Returns: namespace with all parsed arguments from all provided ParserBase subclasses
+        """
+        return cls._parse_args(parser_classes, description, arguments, allow_unkowns=True, **kwargs)
```
thomas-bc: This function is called `parse_known_args` yet uses `allow_unkowns=True` which is a little confusing. Should the keyword name be refined?

thomas-bc (approved):

PR #211, opened 2025-06-10
InT to load a deployment specific config file and provide necessary mapping

ekswang:
| | |
|:---|:---|
|**_Related Issue(s)_**| [3213](https://github.com/nasa/fprime/issues/3213) |

---
## Change Description

Adding a feature to pytest to load a deployment configuration file with a list of deployment configurations. This configuration will list the instance names of F Prime components and the file path to parameter db.
 
- [x] component instance name: get_mnemonic()
- [x] file path to parameter db: get_prm_db_path()

API change:
- [x] get_deployment() - updated to use metadata info in the loaded dictionary rather than using dictionary path

## Rationale

Component instance names can be different across different deployments. As we start developing InT scripts for each component, we'll need to implement them while being deployment agnostic. The InT scripts will utilize the configuration JSON file to know component's instance name and file path to parameter db. This is different than the current approach to hard-coding the instance name in the InT script.

Ran pytest to check the loading of the the deployment configuration file using defined `--deployment-config` flag. Additional tests were performed using this feature as part of InT testing.

## Future Work

N/A

### DIFF
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -31,7 +31,7 @@ class IntegrationTestAPI(DataHandler):
 
     NOW = "NOW"
 
-    def __init__(self, pipeline, deployment_config, logpath=None, fsw_order=True):
+    def __init__(self, pipeline, deployment_config=None, logpath=None, fsw_order=True):
         """
         Initializes API: constructs and registers test histories.
         Args:
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -239,7 +239,7 @@ class IntegrationTestAPI(DataHandler):
         try:
             with open(dictionary, 'r') as file:
                 data = json.load(file)
-            deployment = data['metadata'].get("deploymentName")
+            return data['metadata'].get("deploymentName")
         except FileNotFoundError:
             msg = f"Error: File not found at path: {dictionary}"
             self.__log(msg, TestLogger.YELLOW)
@@ -252,7 +252,6 @@ class IntegrationTestAPI(DataHandler):
             msg = f"An unexpected error occurred: {e} is an unknown key"
             self.__log(msg, TestLogger.YELLOW)
             return None
-        return deployment
 
     def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
         """
@@ -286,6 +285,8 @@ class IntegrationTestAPI(DataHandler):
         """
         if hasattr(self.pipeline, "deployment_config"):
             return self.pipeline.deployment_config
+        else:
+            return None
 
     def load_config_file(self):
         """
@@ -299,6 +300,7 @@ class IntegrationTestAPI(DataHandler):
         try:
             with open(config_file, 'r') as file:
                 result = json.load(file)
+            return result
         except FileNotFoundError:
             msg = f"Error: File not found at path {config_file}"
             self.__log(msg, TestLogger.RED)
@@ -311,15 +313,14 @@ class IntegrationTestAPI(DataHandler):
             msg = f"An unexpected error occurred: {e}"
             self.__log(msg, TestLogger.RED)
             assert False, msg
-        return result
 
-    def getCmdDispName(self, instance=None, name=None):
+    def get_mnemonic(self, comp=None, name=None):
         """
         Get deployment mnemonic of specified item from user-specified deployment
         configuration file.
 
         Args:
-            instance: native component to instance (str), i.e. "<component>.<instance>"
+            comp: qualified name of the component (str), i.e. "<component>.<instance>"
             name: command, channel, or event name (str) [optional]
         Returns:
             deployment mnemonic of specified item (str) or native mnemonic (str) if not found
@@ -328,13 +329,13 @@ class IntegrationTestAPI(DataHandler):
 
         if data:
             try:
-                mnemonic = data[instance]
+                mnemonic = data[comp]
+                return f"{mnemonic}.{name}" if name else f"{mnemonic}"
             except KeyError:
-                self.__log(f"Error: {instance} not found", TestLogger.YELLOW)
-                return f"{instance}.{name}" if name else f"{instance}"
-            return f"{mnemonic}.{name}" if name else f"{mnemonic}"
+                self.__log(f"Error: {comp} not found", TestLogger.YELLOW)
+                return f"{comp}.{name}" if name else f"{comp}"
 
-    def getPrmDbFilename(self):
+    def get_prm_db_path(self) -> str:
         """
         Get file path to parameter db from user-specified deployment configuration file.
 
@@ -345,7 +346,7 @@ class IntegrationTestAPI(DataHandler):
 
         if data:
             try:
-                return data["Svc.PrmDb.filepath"]
+                return data["Svc.PrmDb.filename"]
             except KeyError:
                 return None
 
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -346,7 +346,13 @@ class IntegrationTestAPI(DataHandler):
 
         if data:
             try:
-                return data["Svc.PrmDb.filename"]
+                filepath = data["Svc.PrmDb.filename"]
+                if filepath.startswith('/'):
+                    return filepath
+                else:
+                    msg = f"Error: {filepath} did not start with a forward slash"
+                    self.__log(msg, TestLogger.RED)
+                    assert False, msg
             except KeyError:
                 return None
 
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -11,6 +11,7 @@ import signal
 import time
 from pathlib import Path
 import shutil
+import json
 
 from fprime.common.models.serialize.time_type import TimeType
 
@@ -228,11 +229,30 @@ class IntegrationTestAPI(DataHandler):
 
     def get_deployment(self):
         """
-        Get the deployment of the target using the loaded FSW dictionary path
+        Get the deployment of the target using the loaded FSW dictionary.
+
         Returns:
-            The name of the deployment (str)
+            The name of the deployment (str) or None if not found
         """
-        return Path(self.pipeline.dictionary_path).parent.parent.name
+        dictionary = str(self.pipeline.dictionary_path)
+
+        try:
+            with open(dictionary, 'r') as file:
+                data = json.load(file)
+            deployment = data['metadata'].get("deploymentName")
+        except FileNotFoundError:
+            msg = f"Error: File not found at path: {dictionary}"
+            self.__log(msg, TestLogger.YELLOW)
+            return None
+        except json.JSONDecodeError as e:
+            msg = f"Error decoding JSON: {e}"
+            self.__log(msg, TestLogger.YELLOW)
+            return None
+        except Exception as e:
+            msg = f"An unexpected error occurred: {e} is an unknown key"
+            self.__log(msg, TestLogger.YELLOW)
+            return None
+        return deployment
 
     def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
         """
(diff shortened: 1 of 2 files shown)


src/fprime_gds/common/testing_fw/api.py
```diff
@@ -257,6 +277,78 @@ def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
...
+        Get file path to parameter db from user-specified deployment configuration file.
+
+        Returns:
+            file path to parameter db (str) or None if not found
+        """
+        data = self.load_config_file()
+
+        if data:
+            try:
+                return data["Svc.PrmDb.filepath"]
```
thomas-bc: Very nitpicky but is that the key name? 
It's a little misleading, one could think the JSON representation under the hood would be
```json
{
   "Svc" : { 
       "PrmDb": { "filepath" : "/abc/"}, 
        [...] 
    } 
}
```
But I'm assuming it's just
```json
{
    "Svc.PrmDb.filepath" : "/abc/:
}
```
Do we have an example of how a real-world config file would look like? It would be good to reference it here somewhere.

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -257,6 +277,78 @@ def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
...
+            assert False, msg
+        except Exception as e:
+            msg = f"An unexpected error occurred: {e}"
+            self.__log(msg, TestLogger.RED)
+            assert False, msg
+        return result
+
+    def getCmdDispName(self, instance=None, name=None):
+        """
+        Get deployment mnemonic of specified item from user-specified deployment
```
thomas-bc, line 321: Could we rename the function to something like `get_mnemonic` or something?

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -257,6 +277,78 @@ def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
...
+        configuration file.
+
+        Args:
+            instance: native component to instance (str), i.e. "<component>.<instance>"
```
thomas-bc: `<component>.<instance>` I think this is just the qualified name of the component (_not_ the instance) ?

src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -52,6 +52,12 @@ def pytest_addoption(parser):
         action="store_true",
         help="Enable JUnitXML report generation to a specified location"
     )
+    # Add an option to specify a json config file that maps instances
+    parser.addoption(
+        "--deployment-config",
```
thomas-bc, line 57: This functionality is great - we should document it in https://github.com/nasa/fprime/blob/devel/docs/user-manual/gds/gds-test-api-guide.md

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -257,6 +277,78 @@ def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
...
+
+        if data:
+            try:
+                mnemonic = data[instance]
+            except KeyError:
+                self.__log(f"Error: {instance} not found", TestLogger.YELLOW)
+                return f"{instance}.{name}" if name else f"{instance}"
+            return f"{mnemonic}.{name}" if name else f"{mnemonic}"
+
+    def getPrmDbFilename(self):
```
thomas-bc: Nit: 

```suggestion
    def get_prm_db_path(self) -> str:
```

Should this be an absolute path so it works the same regardless of where `pytest` is ran from?

thomas-bc (changes requested):

src/fprime_gds/common/testing_fw/api.py
ekswang: Yes, the JASON representation is the latter. Kevin provided a config file and it looks as follows:

```
{
    "Deployment" : "LedBlinker",
    "Drv.BlockDriver" : "LedBlinker.blockDrv",
    "Svc.ActiveRateGroup" : "LedBlinker.rateGroup1Comp",
    "Svc.CommandDispatcher" : "LedBlinker.cmdDisp",
    "Svc.CmdSequencer" : "LedBlinker.cmdSeq",
    "Svc.FileDownlink" : "LedBlinker.fileDownlink",
    "Svc.FileManager" : "LedBlinker.fileManager",
    "Svc.FileUplink" : "LedBlinker.fileUplink",
    "Ref.PingReceiver" : "LedBlinker.pingRcvr",
    "Svc.ActiveLogger" : "LedBlinker.eventLogger",
    "Svc.TlmChan" : "LedBlinker.tlmSend",
    "Svc.PrmDb" : "LedBlinker.prmDb",
    "Svc.DpCatalog" : "LedBlinker.dpCat",
    "Svc.DpManager" : "LedBlinker.dpMgr",
    "Svc.DpWriter" : "LedBlinker.dpWriter",
    "Svc.ComQueue" : "LedBlinker.comQueue",
    "Ref.TypeDemo" : "LedBlinker.typeDemo",
    "Svc.Health" : "LedBlinker.health",
    "Ref.SignalGen" : "LedBlinker.SG1",
    "Ref.SendBuff" : "LedBlinker.sendBuffComp",
    "Drv.TcpClient" : "LedBlinker.comDriver",
    "Svc.AssertFatalAdapter" : "LedBlinker.fatalAdapter",
    "Svc.FatalHandler" : "LedBlinker.fatalHandler",
    "Svc.BufferManager" : "LedBlinker.commsBufferManager",
    "Svc.PosixTime" : "LedBlinker.posixTime",
    "Svc.RateGroupDriver" : "LedBlinker.rateGroupDriverComp",
    "Ref.RecvBuff" : "LedBlinker.recvBuffComp",
    "Svc.Version" : "LedBlinker.version",
    "Svc.PassiveTextLogger" : "LedBlinker.textLogger",
    "Svc.SystemResources" : "LedBlinker.systemResources",
    "Svc.BufferManager" : "LedBlinker.dpBufferManager",
    "Svc.FrameAccumulator" : "LedBlinker.frameAccumulator",
    "Svc.FprimeDeframer" : "LedBlinker.deframer",
    "Svc.FprimeRouter" : "LedBlinker.fprimeRouter",
    "Svc.FprimeFramer" : "LedBlinker.fprimeFramer",
    "Svc.ComStub" : "LedBlinker.comStub"
}
```
And the suggested filepath is `Svc.PrmDb.filepath`. **

** Correction: Kevin updated the ticket for filepath to be filename instead. I will correct that in the next set of check-in as part of your suggested changes.
`
 "Svc.PrmDb.filename" : "/tmp/prmdb.db"`

src/fprime_gds/common/testing_fw/api.py
ekswang: Yes, I can change the name to get_mnemonic. I was just going with the API/function name mentioned in the [3213](https://github.com/nasa/fprime/issues/3213).

src/fprime_gds/common/testing_fw/api.py
thomas-bc: Since it's supposed to be a full path, I would argue `filepath` might be better than `filename` . But that's up to yall

src/fprime_gds/common/testing_fw/api.py
thomas-bc: If yes, could maybe enforce that it needs to start with `/`

ekswang: Added the enforcement that it needs to start with /. I wasn't planning on checking the set values in the configuration file other than this exception. I am also assuming that the param db file has an absolute path either from generated/dumped db file via command or a specific version that the user wants to use.

src/fprime_gds/common/testing_fw/api.py
ekswang: updated the description.

src/fprime_gds/common/testing_fw/api.py
ekswang: changed function to get_mnemonic.

src/fprime_gds/common/testing_fw/api.py
ekswang: I will leave that Kevin to decide. For now, it's based on the naming convention provided in https://github.com/nasa/fprime/issues/3213.

kevin-f-ortega: Let's go with Thomas' recommendation of `filepath`.

src/fprime_gds/common/testing_fw/api.py
kevin-f-ortega: LGTM

kevin-f-ortega (commented):
Looks good to me. Thank you

kevin-f-ortega (dismissed):

src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -92,6 +102,10 @@ def fprime_test_api_session(request):
         # Build a new pipeline with the parsed and processed arguments
         pipeline = pipeline_parser.pipeline_factory(arg_ns, pipeline)
 
+        # Add an attribute for deployment configuration file to pipeline
+        if request.config.option.deployment_config:
+            pipeline.deployment_config = request.config.option.deployment_config
```
thomas-bc: Why does `deployment_config` need to be added to `pipeline` ? 

Could it be added straight to the `api` object instead? I'm guessing the API object is what will need that config?

ekswang: Adding it to `pipeline` object allows it to be accessible by `IntegrationTestAPI` class.

thomas-bc: Yes, but the IntegrationTestAPI is the class that needs it, not the `pipeline`, no? 
Why not directly put it in the `IntegrationTestAPI` object? It sounds like it conceptually fits in there better than in `pipeline`

thomas-bc: Sorry maybe I'm being a little too pedantic; it's just that StandardPipeline is the class that kind of orchestrates all of the runtime stuff for a working GDS stack. It sounds odd to shove a configuration property of FSW integration tests in there. Whereas `IntegrationTestAPI` is precisely intended for InT.

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -257,6 +276,86 @@ def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
...
+        """
+        data = self.load_config_file()
+
+        if data:
+            try:
+                mnemonic = data[comp]
+                return f"{mnemonic}.{name}" if name else f"{mnemonic}"
+            except KeyError:
+                self.__log(f"Error: {comp} not found", TestLogger.YELLOW)
+                return f"{comp}.{name}" if name else f"{comp}"
```
LeStarch, line 338: Should use this return value if `data` is not set as well.

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -257,6 +276,86 @@ def wait_for_dataflow(self, count=1, channels=None, start=None, timeout=120):
...
+                filepath = data["Svc.PrmDb.filename"]
+                if filepath.startswith('/'):
+                    return filepath
+                else:
+                    msg = f"Error: {filepath} did not start with a forward slash"
+                    self.__log(msg, TestLogger.RED)
+                    assert False, msg
+            except KeyError:
+                return None
+
```
LeStarch, line 364: Should add explicit return of `None` when data is not set.

src/fprime_gds/common/testing_fw/pytest_integration.py
LeStarch: Agreed.  This should not monkey-patch the pipeline object, but rather use the int test API.

LeStarch (changes requested):

src/fprime_gds/common/testing_fw/pytest_integration.py
ekswang: Copied. I will create a separate argument to pass in this to API rather than adding it to pipeline object.

src/fprime_gds/common/testing_fw/api.py
ekswang: Copied. I will set the return to that.

src/fprime_gds/common/testing_fw/api.py
ekswang: Copied. I will add return of None for data not set as well.

LeStarch (approved):

src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -63,6 +69,10 @@ def pytest_configure(config):
     if config.getoption("--gen-junitxml"):
         config.option.xmlpath = Path(config.getoption("--logs")) / config.getoption("--junit-xml-file")
 
+    # Get the file path to the deployment configuration file
+    if config.getoption("--deployment-config"):
+        deployment_config = config.getoption("--deployment-config")
```
LeStarch: Did you mean to load the deployment config file here?  Otherwise this is an unused variable.

LeStarch (changes requested):

src/fprime_gds/common/testing_fw/api.py
```diff
@@ -30,15 +31,17 @@ class IntegrationTestAPI(DataHandler):
 
     NOW = "NOW"
 
-    def __init__(self, pipeline, logpath=None, fsw_order=True):
+    def __init__(self, pipeline, deployment_config, logpath=None, fsw_order=True):
```
thomas-bc: Making deployment_config optional
```suggestion
    def __init__(self, pipeline, deployment_config=None, logpath=None, fsw_order=True):
```

ekswang: Recommendation added.

thomas-bc (approved):

LeStarch (approved):

PR #214, opened 2025-06-12
Adding in default packet set selection

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Select the default packet set if it is the only one available.

### DIFF
--- a/src/fprime_gds/common/pipeline/dictionaries.py
+++ b/src/fprime_gds/common/pipeline/dictionaries.py
@@ -134,6 +134,7 @@ class Dictionaries:
                 names = packet_loader.get_packet_set_names(None)
                 if len(names) == 0:
                     self._packet_dict = None
+                    return
                 elif len(names) > 1:
                     raise Exception("[ERROR] Multiple packet sets, must set --packet-set-name")
                 packet_set_name = names[0]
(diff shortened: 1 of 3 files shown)


src/fprime_gds/common/pipeline/dictionaries.py
```diff
@@ -128,8 +128,15 @@ def load_dictionaries(self, dictionary, packet_spec, packet_set_name):
...
-        if self._metadata["dictionary_type"] == "json" and packet_set_name is not None:
+        if self._metadata["dictionary_type"] == "json":
             packet_loader = fprime_gds.common.loaders.pkt_json_loader.PktJsonLoader(dictionary)
+            if packet_set_name is None:
+                names = packet_loader.get_packet_set_names(None)
+                if len(names) == 0:
+                    self._packet_dict = None
+                elif len(names) > 1:
+                    raise Exception("[ERROR] Multiple packet sets, must set --packet-set-name")
+                packet_set_name = names[0]
```
thomas-bc, line 140: This will still be reached if `len(names) == 0` no? Should it be in an else clause ?

LeStarch: Fixed!

thomas-bc (approved):

PR #215, opened 2025-06-16
Fpy compiler V0.1

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#3615 |

---
## Change Description

A description of the changes contained in the PR.

## Rationale

A rationale for this change. e.g. fixes bug, or most projects need XYZ feature.

## Future Work


zimri-leisher:
@bocchino @LeStarch 
This PR is ready for review

src/fprime_gds/common/fpy/compiler.py
bocchino, line 1: It might be better to call this file codegen.py.

bocchino: Or you could call it bytecode.py if it generates byte code.

src/fprime_gds/common/fpy/compiler.py
```diff
@@ -0,0 +1,1709 @@
...
+    U64Type,
+    U8Type,
+)
+FLOAT_TYPES = (
+    F32Type,
+    F64Type,
+)
+
+
+FppTypeClass = type[FppType]
```
bocchino: Please document what line 135 is intended to mean. It sounds like the intended meaning of the RHS is something like "for all types T such that T is a subtype of FppType, type[T]." I'm not sure what it actually means in Python, or if it even has a defined meaning, since Python type annotations aren't checked by default.

src/fprime_gds/common/fpy/compiler.py
```diff
@@ -0,0 +1,1709 @@
...
+    args: list[tuple[str, FppTypeClass]]
+
+
+@dataclass
+class FpyCmd(FpyCallable):
+    cmd: CmdTemplate
+
+
+@dataclass
+class FpyBuiltin(FpyCallable):
```
bocchino: Maybe you could rename FpyBuiltin to FpyBuiltinFunction.

bocchino: Or maybe FpyMacro.

src/fprime_gds/common/fpy/compiler.py
```diff
@@ -0,0 +1,1709 @@
...
+@dataclass
+class FpyTypeCtor(FpyCallable):
+    type: FppTypeClass
+
+
+@dataclass
+class FieldReference:
+    """a reference to a field/index of an fprime type"""
+
+    parent: "FpyReference"
```
zimri-leisher: this doesn't handle A(1).x
struct A {
   x: U32
}

zimri-leisher: could dispense with a reference to a field, resolve reference to a value

src/fprime_gds/common/fpy/compiler.py
```diff
@@ -0,0 +1,1709 @@
...
+
+    type_ref: AstExpr
+    """the expression denoting the var's type"""
+    type: FppTypeClass | None = None
+    """the resolved type of the variable. None if type unsure at the moment"""
+    sreg_idx: int | None = None
+    """the index of the sreg it is stored in"""
+
+
+FpyNamespace = dict[str, "FpyReference"]
```
bocchino: Commenting this and other similar definitions would be helpful.

zimri-leisher:
@bocchino I have responded to your feedback

src/fprime_gds/common/fpy/codegen.py
```diff
@@ -0,0 +1,1691 @@
...
+
+    type_ref: AstExpr
+    """the expression denoting the var's type"""
+    type: FppTypeClass | None = None
+    """the resolved type of the variable. None if type unsure at the moment"""
+    sreg_idx: int | None = None
+    """the index of the sreg it is stored in"""
+
+
+# a namespace is just a mapping of strings to FpyReferences
```
bocchino: This comment is a bit confusing. I think a namespace is like a symbol table: it maps fully-qualified names of symbols that are in scope to their definitions. I don't quite understand the second sentence. Does "it" refer to a qualified name that is a key in the map?

bocchino (dismissed):
Looks good! I just had one question about a comment.

src/fprime_gds/common/fpy/codegen.py
zimri-leisher: "it" in the second sentence refers to an FpyNamespace

I think I'm thinking about namespaces in a different way than you. What I'm trying to get at here is that we need some way of hierarchically storing named concepts in Fpy. For instance, to represent the command `Ref.cmdDisp.CMD_NO_OP`, I want to store:
```
{
   "Ref": {
         "cmdDisp": {
                "CMD_NO_OP": FpyCmd(...)
          }
    }
}
```

As opposed to just storing it in a "flat" way like:
```
{
"Ref.cmdDisp.CMD_NO_OP": FpyCmd(...)
}
```

In this example, I am calling each dictionary an "FpyNamespace", but maybe there's a better term for it. It's really just a way of hierarchically storing information about the sequence. Perhaps it doesn't need to be called FpyNamespace at all.

bocchino: OK, it sounds like the dictionary maps unqualified symbol names to symbols. I would call that a scope, or a symbol table corresponding to a scope. I think I was thrown off by the second sentence. (1) In a hierarchical symbol table, a key may or may not be a qualifier. It is a qualifier if and only if the symbol that it refers to also defines a scope. If the symbol that it refers to does not define a scope (e.g., a command definition), then it's not a qualifier; it's just a name. (2) I'm not sure why we are referring to values in this context. A symbol may or may not be associated with a value. For example, variables store values and module qualifiers don't. In any event, the structure of the symbol table seems unrelated to whether the symbols store values.

zimri-leisher: Okay, I will rename it to an "FpyScope"

zimri-leisher:
okay @bocchino , I believe your comment is resolved

bocchino (dismissed):
Looks good!

zimri-leisher:
Unfortunately I cannot seem to fix the spelling check. @LeStarch @thomas-bc  would either of you be able to help here?

zimri-leisher:
@bocchino  sorry to bother, but your review was removed due to some changes. Can you reapprove, and then merge?

thomas-bc (approved):

PR #231, opened 2025-08-20
Fpy compiler 0.2

### LINKED ISSUE #3865: FpySequencer accept int literal for float
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**|  fpyc|
---
## Feature Description

Integer literals should probably be allowed to be converted to floats implicitly. This is more Pythonic and saves people from typing boilerplate `.0`

### LINKED ISSUE #3877: FpySequencer non-constant arguments to functions
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| fpyc, FpySequencer |
---
## Feature Description

Allow expressions with non-constant values to be passed in to functions.
For instance:
```
var1: U32 = 123
sleep(var1, 0)

var2: bool = False
exit(var2)
```

### LINKED ISSUE #4058: FpySequencer error when undefined var
| | |
|:---|:---|
|**_F´ Version_**| 4.0.0 |
|**_Affected Component_**| FpySequencer |
---
## Problem Description

Better error message when there's a random string on a line

Image

### LINKED ISSUE #4077: fpyc remove negative sign from int syntax
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| fpyc |
---
## Feature Description

The negative sign should not be syntactically part of an integer or float literal. Rather, it should be a negation operator and constant folding should be applied to negate the integer

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#4051 |

---
## Change Description

User tutorial in `src/fprime_gds/common/fpy/README.md`
Work in progress language spec in `src/fprime_gds/common/fpy/SPEC.md`

* Stack based architecture! Everything goes on the stack. 
  * The stack currently only has one frame because we don't support function calling. 
  * The frame starts with a local variable array, then on top of that lives the operand stack
* Variable arguments to everything! commands, directives, macros etc
* Simplified syntax tree for operators
* Check use before declare vars
* Well-defined (and documented) (and consistently applied) type conversion rules
* Comparing complex types for equality
* Return status from commands
* Unary plus and minus operators
* A Python model of the FpySequencer, so I can develop new directives and functionality quite quickly, and test it without needing an fprime deployment.
* 900 automatically generated tests of various arithmetic cases
* 120+ manually written test sequences

Closes nasa/fprime#3865
Closes nasa/fprime#3877
Closes nasa/fprime#4077
Closes nasa/fprime#4058


zimri-leisher:
@bocchino @LeStarch I'm eager to start the review process for the latest compiler. However, there's still some work in progress documentation.  I think it would be best if you took a look at just the compiler itself for now: `codegen.py`

src/fprime_gds/common/fpy/bytecode/parser.py
```diff
@@ -0,0 +1,236 @@
...
+    dest: str | int
+
+
+@dataclass
+class AstWaitRelCall(Ast):
+    seconds: int
+    useconds: int
+
+
+@dataclass
```
Kronos3: Might be better to use `@dataclass(frozen=True)` for AST nodes

src/fprime_gds/common/fpy/codegen.py
```diff
@@ -779,118 +865,225 @@ def visit_AstAssign(self, node: AstAssign, state: CompileState):
...
+
+        if issubclass(type, FloatType) and issubclass(to_type, IntegerType):
+            # cannot convert float to int (i.e. don't allow it)
+            return False
+
+        # otherwise we're good
+        return True
+
+    def pick_intermediate_type(
+        self, arg_types: list[FppTypeClass], op: str
```
Kronos3: I think this `op` should be a union of `UnaryOp` and `BinaryOp` enums. You can still make the enum values `str` but I think it's too permissive and leaves room for issues.

Kronos3 (commented):
This looks great! I added some minor code cleanup comments.

On a broader note, the current codegen strategy seems a little clunky. From what I understand, each pass will map a node to a set of directives and then its parent node will just append from there.

I'm curious why you chose this pattern instead of following the LLVM `IRBuilder` pattern where expressions can add instructions to a block and return `Value` and statements can add blocks and perform the control flow. Seems like a builder pattern may simplify much of your code and allow you run passes over IR blocks. Again, not saying to change anything I'm just curious to hear your thoughts.

zimri-leisher:
> This looks great! I added some minor code cleanup comments.
> 
> On a broader note, the current codegen strategy seems a little clunky. From what I understand, each pass will map a node to a set of directives and then its parent node will just append from there.
> 
> I'm curious why you chose this pattern instead of following the LLVM `IRBuilder` pattern where expressions can add instructions to a block and return `Value` and statements can add blocks and perform the control flow. Seems like a builder pattern may simplify much of your code and allow you run passes over IR blocks. Again, not saying to change anything I'm just curious to hear your thoughts.

Oh, honestly I just didn't know how anyone else did this so this is what I came up with. The LLVM strategy seems good, I will likely implement something like this for the next update.

If possible, I'd prefer to merge this clunky strategy now and make it better iteratively.

src/fprime_gds/common/fpy/bytecode/parser.py
zimri-leisher: this would be a good call but I do actually modify them, I give each node an `id` field after construction, so i can hash them by identity more easily. i could probably figure out a way around this but it's a pain

src/fprime_gds/common/fpy/codegen.py
zimri-leisher: oh okay good idea, i should add an enum type yeah

src/fprime_gds/common/fpy/bytecode/grammar.lark
```diff
@@ -0,0 +1,79 @@
...
+STRING: /("(?!"").*?(?<!\\)(\\\\)*?"|'(?!'').*?(?<!\\)(\\\\)*?')/i
+
+_SPECIAL_DEC: "0".."9"        ("_"?  "0".."9"                       )*
+DEC_NUMBER:   "1".."9"        ("_"?  "0".."9"                       )*
+          |   "0"             ("_"?  "0"                            )* /(?![1-9])/
+HEX_NUMBER.2: "0" ("x" | "X") ("_"? ("0".."9" | "a".."f" | "A".."F"))+
+OCT_NUMBER.2: "0" ("o" | "O") ("_"?  "0".."7"                       )+
+BIN_NUMBER.2: "0" ("b" | "B") ("_"?  "0".."1"                       )+
+
+DECIMAL: "." _SPECIAL_DEC | _SPECIAL_DEC "." _SPECIAL_DEC?
```
bocchino: These rules seem inconsistent with the rules for numeric literals in the high-level language. Should there be one set of rules for both? Also, why make `-` part of the numeric literal? It seems cleaner to rely on unary minus for negation.

src/fprime_gds/common/fpy/grammar.lark
```diff
@@ -76,13 +99,25 @@ NAME: /[^\W\d]\w*/
...
+NOT_OP: "not"
 COMPARISON_OP: ">" | "<" | "<=" | ">=" | "==" | "!="
 
 STRING: /("(?!"").*?(?<!\\)(\\\\)*?"|'(?!'').*?(?<!\\)(\\\\)*?')/i
 
-_SPECIAL_DEC: "0".."9"        ("_"?  "0".."9"                       )*
 DEC_NUMBER:   "-"? "1".."9"        ("_"?  "0".."9"                       )*
           |   "-"? "0"             ("_"?  "0"                            )* /(?![1-9])/
 
+_SPECIAL_DEC: "0".."9"        ("_"?  "0".."9"                       )*
```
bocchino, line 119: Can we remove `-` from the syntax of integer literals and use unary minus instead? Or is that future work?

zimri-leisher: Let's keep it this way for now and remove it when I add constant folding. Otherwise, -1 would emit three instructions: push a 1, push a negative 1, and multiply.

src/fprime_gds/common/fpy/bytecode/grammar.lark
zimri-leisher: Ah, yeah the "assembly" grammar is currently not used. I will remove it from this PR

zimri-leisher:
@bocchino comments have been responded to

src/fprime_gds/common/fpy/grammar.lark
bocchino: OK, but unary minus is allowed, right? So if a user writes `- 1` it will generate those instructions. It seems brittle to make an optimization like this depend on spacing. I think it would be better to be consistent, and just accept the fact that the code is unoptimized for now.

zimri-leisher: good point. Okay, I'll make this change

bocchino: Sounds good. I think the docs need to be updated also.

bocchino (dismissed):
This looks good to me! Before we merge it, the CI failures have to be fixed.

zimri-leisher:
@bocchino all checks are passing, please re approve and merge when ready

bocchino (approved):
Looks good!

PR #242, opened 2025-09-12
Fix a small bug in codegen for StackCmd

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#4130 |

---
## Change Description
This change fixes a bug in the codegen for the StackCmd directive. Previously, the wrong size for arguments was calculated when there was more than one arg, leading to a format error for such cmds when called with variable args. This is now fixed and a test was added to catch this and similar errors.


zimri-leisher:
@LeStarch  can you 1) give me permission to request reviews and 2) add Andrei to this PR? We found this bug together so he should be able to review it.

LeStarch (approved):

PR #245, opened 2025-09-19
Fpy assembler

zimri-leisher:
| | |
|:---|:---|
|**_Related Issue(s)_**|  nasa/fprime#4064|

---
## Change Description

* Fpy bytecode human-readable grammar, parser and assembler. This lets you write Fpy bytecode directly, without having to go through the compiler.
* Fpy binary disassembler. This turns a .bin file into a human-readable bytecode file, letting you inspect previously compiled sequences
* Compiler bytecode output option. This lets you output human readable bytecode from high-level Fpy
* Looped the assembler and disassembler into the test sequences. Every single test seq gets compiled, then disassembled, then assembled, then run.


LeStarch (commented):
Skimmed through this, looks reasonable. 

I wonder if this product should be moved into a separate repository.  There is a lot of code, multiple executables, and it is adding new dependencies. Thoughts?

zimri-leisher:
Do you mean specifically the assembler/disassembler? As far as I can tell, it's not adding any new dependencies, but yes I guess it does add some more executables. If you could enumerate more specifically your concerns, I think it could be moved into a separate repo. Although we'd have to work on the workflow--how it would get published/reviewed etc.

zimri-leisher:
@LeStarch happy to continue discussing this, but I would like to resolve this discussion today or tomorrow because I'd like collaborators to be able to use this tool. If you're pressed for time, I'd prefer just merging this PR and we can discuss another repo during the next meeting.

LeStarch (approved):

PR #249, opened 2025-09-30
Add TM Fixed Frame Size CLI option

Willmac16:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

makes TM_FIXED_FRAME_SIZE a CLI argument so users can reconfigure GDS to match the value specified in ComCfg.fpp (and actually rx packets).

## Rationale

Fprime GDS does not work with TM Fixed Frame Values other than 1024 until this change is applied.

Decide if random testing is needed on the frame size & if any more checks are needed to validate the argument.

## Future Work

N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 docs/doxy/**
 *.log
 .vscode
+*.swp
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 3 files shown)


thomas-bc (approved):
LGTM! Good catch. Note that in the future this (among other configs in `ComCfg.fpp` such as spacecraft ID, APIDs etc.) will be passed through the dictionary.
See https://github.com/nasa/fprime/issues/3664

Willmac16:
> LGTM! Good catch. Note that in the future this (among other configs in `ComCfg.fpp` such as spacecraft ID, APIDs etc.) will be passed through the dictionary. See [nasa/fprime#3664](https://github.com/nasa/fprime/issues/3664)

Oh Awesome! That'll make our GSW engineers much happier

LeStarch:
This is awesome!  I ran into this problem yesterday, went to fix it, and found it done!  Thanks!

PR #252, opened 2025-10-10
Fixed AttributeError when using dat subparser

knockout101:
| | |
|:---|:---|
|**_Related Issue(s)_**| n |

---
## Change Description

Added a check for the attribute `save` when using the dat subparser.

## Rationale

The current state of the `fprime-prm-write dat` command breaks because there is no args.save. The way the source code is written this is a requirement. This change fixes this issue without changing the fundamental functionality. 

Attempt to run `fprime-prm-write dat` with the intended arguments described in --help.

### DIFF
--- a/src/fprime_gds/common/tools/params.py
+++ b/src/fprime_gds/common/tools/params.py
@@ -211,6 +211,10 @@ def main():
         output_path = args.json_file.with_suffix("." + output_format)
     else:
         output_path = args.output
+
+    # when using dat need a save attribute
+    if not hasattr(args, "save"):
+        args.save = False
     
     convert_json(args.json_file, args.dictionary, output_path, output_format, args.defaults, args.save)
 


LeStarch:
Shouldn't this be done via a default value on this line:

```diff
-json_to_seq.add_argument("--save", action="store_true", help="Whether or not to include the PRM_SAVE cmd in the output")
+json_to_seq.add_argument("--save", action="store_true", help="Whether or not to include the PRM_SAVE cmd in the output", default=False)
```

knockout101:
I tested this idea, but it does not work.

`json_to_seq` is one of the two subparsers within the `main()`. However, the issue is originating from the `json_to_dat` subparser. The `json_to_dat` subparser doesn't have a --save flag and thus `args` object has no attribute save. My "straight forward" approach was to simply add the attribute save to `args` if it does not already have the attribute, so it does not overwrite a good value and will set it if it is missing.

This solution alleviates the issue without an overhaul of the parsing system or adding an unnecessary --save flag to the `json_to_dat` subparser when it is not a useful attribute in that workflow.

LeStarch:
That makes sense.  @zimri-leisher would you review this change?

zimri-leisher (approved):
I think this is the right solution. Apologies for not catching this when originally written.

LeStarch (approved):

PR #256, opened 2025-10-16
Set the default configuration file path to $FPRIME_GDS_CONFIG_PATH when set

ianbrault:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4204 |

---
## Change Description

This modifies the `run_deployment` executable to check if the FPRIME_GDS_CONFIG_PATH environment variable is set and, if so, sets the `ConfigDrivenParser` default path to be this value so it becomes the default value for the `-c`/`--config` command-line argument.

## Rationale

We deliver our project in a containerized environment where the `fprime-gds.yml` file cannot always be installed in the end user's current directory. It is more convenient to have an environment variable configured to point to the configuration file instead of requiring the end user to pass `-c <long/path>` every time they launch the GDS.

Set the `FPRIME_GDS_CONFIG_PATH` environment variable and verify that the config file from that path is loaded successfully.

## Future Work

N/A

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -390,7 +390,7 @@ class ConfigDrivenParser(ParserBase):
                 "required": False,
                 "default": self.DEFAULT_CONFIGURATION_PATH,
                 "type": Path,
-                "help": f"Argument configuration file path.",
+                "help": "Argument configuration file path. [default: %(default)s]",
             }
         }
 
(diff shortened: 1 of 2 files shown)


thomas-bc (approved):
Looks correct to me, assuming this has been tested

ianbrault:
@thomas-bc yes I have tested it

thomas-bc:
Thanks a lot @ianbrault  !

PR #259, opened 2025-10-30
Pull CCSDS config from dictionary if available

### LINKED ISSUE #3664: fprime-gds: Read in CCSDS constants from dictionary
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Once CCSDS constants are in the JSON dictionary (see https://github.com/nasa/fpp/issues/727), we'll want to update the GDS use the constants defined in the dictionary. In the event they are not in the dictionary, use the default.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/3664 |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3664

- Refactor the ConfigManager logic a little bit to make it easier to work with and purging of the ConfigParser (INI files) that was unused
- Refactor `DataDescType` and `ApidType` (almost aliases) to be Enum-like but pulled from the ConfigManager at runtime
- Add functionality to FramerDeframer to pull config values at instantiation


src/fprime_gds/common/distributor/distributor.py
```diff
@@ -52,12 +52,6 @@ def __init__(self, config=None):
 
         # Internal buffer for un distributed data
         self.__buf = bytearray(b"")
-        # Setup key framing
```
LeStarch, line 55: Why was this removed?  This code allows us to handle processing of data found from `Svc.ComLogger` files that include an internal framing key.

src/fprime_gds/common/distributor/distributor.py
```diff
@@ -95,22 +89,6 @@ def parse_into_raw_msgs_api(self, data):
         raw_msgs = []
         # Keep parsing and then break when you can't parse no more
         while True:
-            # Search data looking for key-frame
```
LeStarch, line 98: Same here.

test/fprime_gds/common/communication/ccsds/test_space_packet.py
```diff
@@ -8,8 +8,8 @@ def framer_deframer():
 
 def test_frame_valid_data(framer_deframer):
     """Test framing valid data (if applicable)."""
-    # Prefix with 4 bytes corresponding to the APID
-    data = b"\x00\x00\x00\x01test_payload"
+    # Prefix with 2 bytes corresponding to the ApidType=DataDescType
```
LeStarch: Does this imply tests will break if `DataDescType` configuration changes?

src/fprime_gds/common/communication/ccsds/space_data_link.py
```diff
@@ -30,12 +32,31 @@ class SpaceDataLinkFramerDeframer(FramerDeframer):
...
+            dict_scid = ConfigManager.get_instance().get_constant("ComCfg.SpacecraftId")
+        except ConfigBadTypeException:
+            pass # Config value not found, move on
+        try:
+            dict_frame_size = ConfigManager.get_instance().get_constant("ComCfg.TmFrameFixedSize")
+        except ConfigBadTypeException:
+            pass # Config value not found, move on
+
         self.sequence_number = 0
+        self.vcid = vcid
```
LeStarch, line 69: Why is vcid not also in dictionary configuration?

src/fprime_gds/common/utils/data_desc_type.py
```diff
@@ -5,30 +5,106 @@
...
+            if apid_type is not None and hasattr(apid_type, "ENUM_DICT"):
+                cls.UNDERLYING_ENUM = Enum(
+                    "DataDescType",
+                    apid_type.ENUM_DICT,
+                )
+                cls.TOKEN_TYPE = enum_type.REPRESENTATION_TYPE_MAP[apid_type.REP_TYPE]
+            else:
+                print(
+                    f"[WARNING] Dictionary does not contain a {cls.ENUM_TYPE_NAME} "
+                    "enumeration. Using default width for Packet Descriptors."
```
thomas-bc:
```suggestion
                    "enumeration. Using default values for Packet Descriptors and APIDs."
```

src/fprime_gds/common/communication/ccsds/space_data_link.py
```diff
@@ -30,12 +32,31 @@ class SpaceDataLinkFramerDeframer(FramerDeframer):
     )
     CRC_CALCULATOR = crc.Calculator(CRC_CCITT_CONFIG)
 
+    # For backwards compatibility if not found in dictionary (loaded by ConfigManager)
+    FALLBACK_SCID = 0x44
+    FALLBACK_FRAME_SIZE = 1024
+
+
     def __init__(self, scid, vcid, frame_size):
```
LeStarch, line 40: Frame size, SCID, and VCID seem like they should only source from the dictionary, no?  These constants must be in-line with FSW.

LeStarch: Perhaps it is an error to specify them on the command line and through the dictionary with inconsistent values.

src/fprime_gds/common/loaders/constant_json_loader.py
```diff
@@ -0,0 +1,52 @@
...
+        for constant in self.json_dict[self.CONSTANTS_FIELD]:
+            try:
+                name_dict[constant["qualifiedName"]] = constant["value"]
+            except KeyError as e:
+                raise GdsDictionaryParsingException(
+                    f"{str(e)} key missing from Constant dictionary entry: {str(constant)}"
+                )
+
+        return (
+            dict(sorted(id_dict.items())),
```
LeStarch: ID dictionary doesn't store anything.

src/fprime_gds/common/loaders/constant_json_loader.py
```diff
@@ -0,0 +1,52 @@
...
+            (id_dict, name_dict). The keys should be the type id and
+            name fields respectively and the values should be type name
+            strings. Note: An empty id dictionary is returned since there
+            are no id fields in the Fw type alias JSON dictionary entries.
+        """
+        id_dict = {}
+        name_dict = {}
+
+        if self.CONSTANTS_FIELD not in self.json_dict:
+            raise GdsDictionaryParsingException(
```
LeStarch: Does this make it dictionary parsing incompatible with older JSON dictionaries?

LeStarch (changes requested):
The biggest concern is the deleted features of the frame sync key.  The others are just minor questions

src/fprime_gds/common/communication/ccsds/space_data_link.py
thomas-bc: VCID doesn't exist in `ComCfg` because our comms stack by default accepts all VCIDs. The Deframer can be configured to accept only one, but since this isn't the default it didn't make it into `ComCfg`. We could add it.

src/fprime_gds/common/communication/ccsds/space_data_link.py
thomas-bc: Good point, I will make it an error to specify something different than what's found in the dictionary, if found

src/fprime_gds/common/loaders/constant_json_loader.py
thomas-bc: Yes cause constants / typedefs don't have an ID (=opcode). I will return `{}` to make it more explicit. This was copied over.

src/fprime_gds/common/loaders/constant_json_loader.py
thomas-bc: Ah I thought `constants` have always had their field in the dictionary, but seems they've been added later on. Will fix.

test/fprime_gds/common/communication/ccsds/test_space_packet.py
thomas-bc: Changed to use configuration types instead

src/fprime_gds/common/communication/ccsds/space_data_link.py
thomas-bc: The reason we have Spacecraft ID in `ComCfg` but not VCID is because Spacecraft ID is tied to a FSW build, where VCID is more loosely defined by the standard (on purpose). One GDS<=>SC link can communicate over multiple Virtual channels, we just happen to only support one channel in F´ currently. 
A project might choose to implement more complex framers than we do, in which case a single `VcIdValue` in the model doesn't make sense anymore.
The standard does say "TM frame size is fixed for the lifespan of the mission" hence putting it as a single constant in the model.

src/fprime_gds/common/communication/ccsds/space_data_link.py
thomas-bc: If it's an error then there's no reason to keep the option moving forward (other than backwards compatibility with `v4.0.0` exclusively) and I'd argue we should remove it. I kept the option and printing a warning instead, there may be some testing scenarios where overriding could be relevant. Let me know

src/fprime_gds/common/distributor/distributor.py
thomas-bc: Added the code back along with updated config handling

LeStarch (dismissed):

LeStarch (approved):

PR #262, opened 2025-11-18
Use configurable types from FSW dictionary across the board

### LINKED ISSUE #4197: Commands with strings are not parsed properly if FwSizeStoreType is not U16
| | |
|:---|:---|
|**_F´ Version_**| 4.0 |
|**_Affected Component_**|   |
---
## Problem Description

Commands with strings, such as `cmdSeq.CS_RUN`, sent by the GDS are not decoded properly in fprime and result in a FORMAT_ERROR. This is due to the 

## Context / Environment

Note: Can provide full report if necessary, but I'm running F Prime on a system without internet which makes this difficult.
```
Linux
aarch64
Pip packages:
 fprime-tools==3.6.1
 fprime-gds==3.6.1
 fprime-fpp-*=2.3.0
fprime @ v3.6.2
```

## How to Reproduce

1. Take the Ref deployment and update FwSizeStoreType to `U64`
2. Send a `CMD_NO_OP_STRING` from the GDS
3. A "FORMAT_ERROR" will appear in the events tab and the command will not be run.

Note: I have not actually tried these steps, but they should be sufficient to reproduce

## Expected Behavior

The command is parsed properly

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

- Removes all found usage of hardcoded-width types with types loaded from FSW dictionary.
- Pulls in the entire types module from `fprime-tools` into `fprime-gds`. We should deprecate or remove the types on the `fprime-tools` side. Types were in tools for legacy reason, and only GDS uses them, so no reason to keep them on the other side. This way, GDS doesn't depend on tools, and tools doesn't depend on GDS.

Fix nasa/fprime#4197

## Review recommendations

Changes to the types are shown with this diff-view : https://github.com/nasa/fprime-tools/compare/devel...thomas-bc:fprime-tools:4197-no-merge-for-review

## Future work

Merge https://github.com/nasa/fprime-tools/pull/296 before next release so as not to define types in both places.


src/fprime_gds/common/models/serialize/string_type.py
```diff
@@ -0,0 +1,114 @@
...
+        # If val is never set then it is init exception...
+        if self.val is None:
+            raise NotInitializedException(type(self))
+        # Check string size before serializing
+        if self.MAX_LENGTH is not None and len(self.val) > self.MAX_LENGTH:
+            raise StringSizeException(len(self.val), self.MAX_LENGTH)
+        # Pack the string size first then return the encoded data buffer
+        return struct.pack(
+            ConfigManager().get_type("FwSizeStoreType").get_serialize_format(),
+            len(self.val),
```
thomas-bc: This should be 

```suggestion
            len(self.val.encode(DATA_ENCODING)),
```

src/fprime_gds/common/models/serialize/string_type.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def deserialize(self, data, offset):
+        """
+        Deserializes a string from the given data buffer.
+        """
+        FwSizeStoreType: type[IntegerType] = ConfigManager().get_type("FwSizeStoreType")  # type: ignore
+        try:
+            val_size = struct.unpack_from(
+                FwSizeStoreType.get_serialize_format(), data, offset
+            )[0]
```
thomas-bc: Should look like:

```

            val_size, = struct.unpack_from(
                FwSizeStoreType.get_serialize_format(), data, offset
            )
```

src/fprime_gds/common/models/serialize/time_type.py
```diff
@@ -0,0 +1,486 @@
...
+            An EnumType instance representing the TimeBase type
+        """
+        return ConfigManager().get_type("TimeBase")(enum_constant)  # type: ignore
+
+    def __init__(self, time_base=0, time_context=0, seconds=0, useconds=0):
+        """
+        Constructor
+
+        Args
+            time_base (int): Time base index for the time tag. Must be a valid
```
thomas-bc: change time_base arg to be TimeBase (EnumType) object rather than

src/fprime_gds/common/utils/config_manager.py
```diff
@@ -79,51 +86,47 @@ def get_instance():
             ConfigManager.__instance = ConfigManager()
         return ConfigManager.__instance
```
thomas-bc: Just `return ConfigManager()`

src/fprime_gds/common/encoders/file_encoder.py
```diff
@@ -43,11 +43,11 @@
 
 import struct
 
-from fprime.common.models.serialize.numerical_types import U32Type
+from fprime_gds.common.models.serialize.numerical_types import U32Type
 from fprime.constants import DATA_ENCODING
```
thomas-bc: Ah! This is using the old `DATA_ENCODING` definition. Need to move to the new definition in fprime-gds

src/fprime_gds/common/encoders/seq_writer.py
```diff
@@ -59,13 +54,13 @@ def __time_tag(cmd_obj):
 
         def __descriptor(cmd_obj):
             # subtract 1 from the value because enum34 enums start at 1, and this can't be changed
+            # REVIEW NOTE: I haven't been to track down if this is configurable or not
```
thomas-bc: https://github.com/nasa/fprime/blob/402c0317b6297d7f436477132f3dc29f152d27a9/Svc/CmdSequencer/FPrimeSequence.cpp#L252

Hardcoded to U8

src/fprime_gds/common/models/serialize/time_type.py
```diff
@@ -0,0 +1,486 @@
...
+
+@date Updated July 22, 2019
+@author Kevin C Oran (kevin.c.oran@jpl.nasa.gov)
+
+@bug No known bugs
+"""
+
+import datetime
+import math
+from enum import Enum
```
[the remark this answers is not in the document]
thomas-bc: Clean up all unused imports

thomas-bc:
@zimri-leisher just putting this PR on your radar. from your side nothing should impact you very much, just the imports are changing from `fprime.common.serialize` to `fprime_gds.common.serialize`. But you may also encounter some merge conflicts. This also speaks to the fact that we should likely pull `fpyc` out of `fprime-gds` and into its own product.

src/fprime_gds/common/models/serialize/time_type.py
thomas-bc: Done - added type checking to allow for either an `int` or a `TimeBase` object because `fpyc` parses it as an int and I don't want to go down that rabbit hole in this PR

LeStarch (approved):

PR #267, opened 2025-12-08
Fall back to defaults for project-unspecified fields in fprime-gds front-end config

pjromano:
| | |
|:---|:---|
|**_Related Issue(s)_**| [fprime #4468](https://github.com/nasa/fprime/issues/4468) |

---
Note: This PR builds on commits from PR #266 (the timeToStringFn config field added by that PR is also included in this change), so PR #266 should be merged first.

This change isn't necessary to resolve ticket [fprime #4468](https://github.com/nasa/fprime/issues/4468), but review updates to config.js in #266 may feed forward to the new config_init.js in this PR.

## Change Description

For configuration of the GDS front-end, this change allows projects to assign only those config fields they wish to set, while leaving the other fields at fprime-gds default values. Note that this modifies the method with which the project specifies their configuration for the GDS front-end, by expecting the project to provide a function that modifies the config object provided in argument, rather than (as-was:) declare and construct the config object themselves.

Documentation previously-missing is added in fprime_gds/flask/static/js/config_init.js, describing the process to establish project-specific front-end config values without forking fprime-gds.

## Rationale

As implemented prior to this PR, a project must create a file that creates and exports the config object with all fields that fprime-gds reads and expects to exist. Once a project decides to point to their own JavaScript file defining the global config object, they might not specify all the fields that the fprime-gds is using/expecting, leading to strange behavior in the front-end operating on undefined values. This could easily come about if the fprime-gds front-end is updated with new config fields and a project updates fprime-gds without recognizing the need to add new fields to their config, leading to issues likely-difficult to trace back to missing fields in their config.

This change allows projects to assign values only to those front-end config fields they are interested in setting, while fprime-gds takes responsibility (instead of project) for initializing any other fields to default values.

I manually tested running through the process for setting configuration from project files, as newly documented at the top of fprime_gds/flask/static/js/config_init.js. Confirmed that values assigned by the project configuration file take place in the GDS GUI, and properties unassigned by the project use the default value (checked the timeToStringFn, dataPollIntervalsMs, and projectName properties).

I also confirmed that leaving the FP_FLASK_SETTINGS environment variable unset results in the default front-end config values (checked the effects of the same 3 properties).

## Future Work

Projects may want to reuse code for handling their custom time bases, or otherwise have the back-end serve files referenced by the config from their project files outside the fprime-gds repo. To allow projects' configuration script to reuse external JavaScript functions by importing them into the configuration script, we could add a new configurable route in the Flask application (configurable similar to "/js/config.js" in fprime_gds/flask/app.py) allowing the project to specify a directory on the back-end host within which project-specified resources can be served under a particular URL, e.g.: URL "/project/<path:path>" would read files from a configurable base path followed by <path:path> on the back-end host's filesystem.

### DIFF
--- a/src/fprime_gds/flask/static/addons/commanding/command-history.js
+++ b/src/fprime_gds/flask/static/addons/commanding/command-history.js
@@ -71,7 +71,7 @@ Vue.component("command-history", {
             for (let i = 0; i < command_copy.args.length; i++) {
                 command_argument_assignment_helper(command_copy.args[i], item.args[i]);
             }
-            return [timeToString(item.datetime || item.time), command_display_string(command_copy)];
+            return [timeToString(item.time), command_display_string(command_copy)];
         },
         /**
          * Take the given item and converting it to a unique key by merging the id and time together with a prefix
--- a/src/fprime_gds/flask/static/js/config.js
+++ b/src/fprime_gds/flask/static/js/config.js
@@ -23,6 +23,8 @@ export let config = {
     },
     // Summary counter fields containing object of field: bootstrap class
     summaryFields: {"WARNING_HI": "warning", "FATAL": "danger", "GDS_Errors": "danger"},
+    // Function to use for formatting timestamps in the tables. null provides default formatting
+    timeToStringFn: null,
 
     // Dashboards are a security vulnerability in that users are uploading artifacts that trigger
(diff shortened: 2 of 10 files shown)


thomas-bc (approved):
Tested and works as expected! Thanks a lot for the detailed docs about it

PR #270, opened 2025-12-16
Load default values in type constructs and pass through to GUI inputs

### LINKED ISSUE #4481: JsonLoader parse default values of types
| | |
|:---|:---|
|**_F´ Version_**| 4.1.0 |
|**_Affected Component_**| fprime-tools JsonLoader |
---
## Feature Description

The JsonLoader should parse the default value of types. Right now it just discards that.

## Rationale

This would be useful for tools which depend on the default value of struct fields.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/4481 |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/4481

## Rationale

This is model information and is valuable to the GDS and downstream tools.


src/fprime_gds/common/loaders/json_loader.py
```diff
@@ -251,9 +254,9 @@ def construct_serializable_type(
                     ),
                 )
             fmt_str = JsonLoader.preprocess_format_str(
-                member_type_obj.FORMAT if hasattr(member_type_obj, "FORMAT") else "{}"
+                member_dict.get("format", "{}")
             )
-            description = member_type_dict.get("annotation", "")
+            description = member_dict.get("annotation", "")
```
thomas-bc, line 259: This and two lines above are unrelated bugfixes that I found while working in there

src/fprime_gds/flask/static/addons/commanding/arguments.js
```diff
@@ -111,6 +111,11 @@ export function squashify_argument(argument) {
     // Base assignment of the value
     let value = argument.value;
 
+    // If already a number, return as-is
+    if (typeof value === "number") {
+        return value;
```
thomas-bc, line 116: This was needed to be able to parse structs' anonymous member-arrays

thomas-bc:
The tests are failing because there are conflict between my modified `RefTopologyDictionary.json` and the one under `fpy`. My belief is that because pytest runs in a single process, it tests `fpy` first and loads its test dictionary, and then the global config keeps these old types and it causes issues when loading a second dictionary. To keep in mind for future tests. Easiest fix is to purge `fpy` from this repo as it has moved: done in https://github.com/nasa/fprime-gds/pull/271

zimri-leisher:
Thanks for working on this! Yup, makes sense to remove fpy first and then do this fix.

LeStarch (approved):

PR #273, opened 2025-12-22
add `--file-uplink-chunk-size argument`

jrpear:
add `--file-uplink-chunk-size argument` to control file uplink chunk size


LeStarch (approved):
Works on my end.

PR #275, opened 2026-01-13
delete file if it exists when the start packet is received

keck-in-space:
| | |
|:---|:---|
|**_Related Issue(s)_**|  NA |
|**_Has Unit Tests (y/n)_**| N |
|**_Documentation Included (y/n)_**| N |

---
## Change Description

We noticed an issue where files with the same name are overwritten, but instead of being truncated, they are appended to from the zeroth index. This resulted in the file on disk (on the target) not matching the file in the fprime-downlinks folder.

This PR is as an attempt at a fix to this issue, with some limited understanding of how GDS works. Please feel free to re-write this patch or treat it as an issue.

## Rationale

This PR prevents files from being in an inconsistent state as it removes a file if it already exists when the START_PACKET is received.

Suggest creating a file in the downlinks folder of a specific length (say 100 bytes of 0s) and then downlinking a file of a shorter length (perhaps 50 bytes of 1s). Both files must have the same name. After issuing the downlink command, the resulting file in the downlinks folder would be half 1s and half 0s, in state that is not consistent with the file on disk (for the target) Should be 50 bytes of 1's in this example.

## Future Work

None come to mind.

### DIFF
--- a/src/fprime_gds/common/files/helpers.py
+++ b/src/fprime_gds/common/files/helpers.py
@@ -141,6 +141,8 @@ class TransmitFile:
         self.__mode = mode
         if mode == TransmitFileState.WRITE:
             filepath = self.__destination
+            if os.path.exists(filepath):
+                os.remove(filepath)
             Path(filepath).touch(exist_ok=True)
             filemode = "rb+"
         else:


thomas-bc:
Yeah this has been brought up in the past. (https://github.com/nasa/fprime/issues/2730)
The issue is that there's no way to tell what the most correct approach is, as different projects might want different behaviors.
 
> The reason the GDS is currently designed this way is to properly handle partial file downlinks (i.e. Filedownlink.SendPartial). You can do multiple partial downlinks, appending/updating the existing file.
 
Especially with the recent PR that was merged: https://github.com/nasa/fprime-gds/pull/274, the above quote is even-more-true

keck-in-space:
Got it. This is very non-intuitive to me. It is effectively an implicit partial overwrite of a file as part of a downlink request. I would have expected the partial downlink command to only give me the segment of the file I requested, not append into an existing file. Now that I know this, I can work around it by `rm -rf` the dowlinks folder prior to every run of GDS.

If this is the intended design, I would suggest documenting it more clearly in the GDS User's Guide. That way others aren't confused as I was when testing our downlinking and getting an amalgamation of old and new files, with the actual resulting file becoming something that doesn't even exist on the target and in many cases would be corrupt.

thomas-bc:
> I would have expected the partial downlink command to only give me the segment of the file I requested, not append into an existing file

@keck-in-space note that this has also been brought up as a potential solution in https://github.com/nasa/fprime/issues/2730 (timestamped files), so maybe that's really what we want to be doing... I don't have a strong opinion. Maybe it ought to be discussed more.

PR #277, opened 2026-01-21
Updated StandardPipeline.setup() to set dictionary_path

bevinduckett:
`IntegrationTestApi.get_deployment()` uses the `dictionary_path` property of self.pipeline, so `StandardPipeline.setup()`  needs to set that value

| | |
|:---|:---|
|**_Related Issue(s)_**|  none |
|**_Has Unit Tests (y/n)_**|  no |

---
## Change Description

Update StandardPipeline.setup()  to set the object's `dictionary_path` field 

## Rationale

Without this change, `IntegrationTestApi.get_deployment()` always fails 

Add `get_deployment()` to any test using `fprime_test_api`, without this change it will fail with `Error: File not found at path: None`. With this change it'll work 

## Future Work

### DIFF
--- a/src/fprime_gds/common/pipeline/standard.py
+++ b/src/fprime_gds/common/pipeline/standard.py
@@ -45,7 +45,6 @@ class StandardPipeline:
         self.distributor = None
         self.client_socket = None
         self.logger = None
-        self.dictionary_path = None
         self.up_store = None
         self.down_store = None
 
@@ -251,3 +250,8 @@ class StandardPipeline:
     def dictionaries(self):
         """Dictionaries member"""
         return self.__dictionaries
+
+    @property
+    def dictionary_path(self):
+        """Dictionary file path"""
+        return self.dictionaries.dictionary_path if self.dictionaries else None
--- a/src/fprime_gds/common/pipeline/standard.py
+++ b/src/fprime_gds/common/pipeline/standard.py
@@ -89,7 +89,6 @@ class StandardPipeline:
                 f"{file_store} is not writable. Fix permissions or change storage directory with --file-storage-directory."
             )
         self.__dictionaries = dictionaries
-        self.dictionary_path = getattr(self.dictionaries,"_dictionary_path")
         self.coders.setup_coders(
             self.dictionaries, self.distributor, self.client_socket
         )
--- a/src/fprime_gds/common/pipeline/standard.py
+++ b/src/fprime_gds/common/pipeline/standard.py
@@ -90,6 +90,7 @@ class StandardPipeline:
                 f"{file_store} is not writable. Fix permissions or change storage directory with --file-storage-directory."
             )
         self.__dictionaries = dictionaries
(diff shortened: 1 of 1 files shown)


thomas-bc:
> Without this change, IntegrationTestApi.get_deployment() always fails

@kevin-f-ortega is this function not actively being used?

src/fprime_gds/common/pipeline/standard.py
```diff
@@ -90,6 +90,7 @@ def setup(
                 f"{file_store} is not writable. Fix permissions or change storage directory with --file-storage-directory."
             )
         self.__dictionaries = dictionaries
+        self.dictionary_path = getattr(self.dictionaries,"_dictionary_path")
```
thomas-bc: This should work as well since it's defined as a `@property`, no?
```suggestion
        self.dictionary_path = self.dictionaries.dictionary_path
```

thomas-bc: Technically I'd actually argue the bug is that `dictionary_path` shouldn't be a property of StandardPipeline at all, so the cleanest would be to fix `get_deployment()` function to use:

```diff
+ dictionary = str(self.pipeline.dictionaries.dictionary_path)
- dictionary = str(self.pipeline.dictionary_path)
```

No reason to unnecessarily copy the `dictionary_path` up into StandardPipeline since we already have a reference to the Dictionaries object, which holds `dictionary_path`

bevinduckett: What tipped me towards setting `self.dictionary_path` in `StandardPipeline.setup()` was that `dictionary_path` is defined as one of the attributes of the object in `StandardPipeline.__init__()` and the effort to verify the code-base wide implications of removing that attribute is higher (i.e. setting the attribute/property to a valid value is fully backwards compatible, removing it is not) 
What do you think of instead removing the dictionary_path attribute & adding the below property function to StandardPipeline? 
```Python 
    @property
    def dictionary_path(self):
        """Dictionaries member"""
        return self.dictionaries.dictionary_path if self.dictionaries else None
```

thomas-bc: @bevinduckett I like that idea a lot!

bevinduckett: Committed discussion change

thomas-bc (dismissed):
LGTM.

@kevin-f-ortega please confirm when you get back, I would have expected for example https://github.com/nasa/fprime/blob/a9c3726c451a1530bd3e8c71018192976f4506ca/Svc/CmdSequencer/test/int/test_cmd_sequencer.py#L127 to be running in CI. Is it not the case?

EDIT: Ah - it seems the function is being called, but the result is not being used (only printed out)

src/fprime_gds/common/pipeline/standard.py
```diff
@@ -90,6 +89,7 @@ def setup(
...
```
thomas-bc: I think removing this line should solve CI error

thomas-bc (approved):

PR #278, opened 2026-01-26
Consolidate Data products ground scripts

### LINKED ISSUE #4642: fprime-write-dp dictionary processing requires all fields
| | |
|:---|:---|
|**_F´ Version_**|4.x |
|**_Affected Component_**|  n/a |
---
## Problem Description

When processing the dictionaries for data product definitions, the script requires all optional fields. If the user doesn't specify them all in the component FPP, the script will error out.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.39
Python version: 3.12.3
CMake version: 3.26.0
Pip version: 24.0
Pip packages:
    fprime-tools==4.0.2a1
    fprime-gds==4.0.2a4
    fprime-fpp==3.0.1a1
```

## How to Reproduce

1. Define some data product containers without annotations or default priorities
2. Generate the data product files
3. Attempt to process them. The script will error.

## Expected Behavior

The script should ignore the containers section of the dictionary anyway since there isn't any useful information for decoding, since the containers are typeless. The script should only read the records section and use that to decode the data in the containers.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/4642, implements part of https://github.com/nasa/fprime/issues/4623

- Uses GDS-provided dictionary loaders rather than doing ad-hoc loading
- Does header parsing by using a pre-defined type
- Refactor to common/executable reusable structure

## AI Usage

Windsurf (Claude Sonnet) was heavily used to get started on the common structure, and hand-implementation later on for the logic of decoding data products
All tests were also written by AI, since the tests are about comparing against a reference output, it was pretty handy and I would deem it fairly safe to do.


src/fprime_gds/common/dp/common.py
```diff
@@ -0,0 +1,91 @@
...
+    (for validating checksums).
+    
+    Args:
+        data: Bytes to calculate checksum for
+        init_value: Initial CRC value (default: 0)
+        
+    Returns:
+        Calculated CRC32 checksum as 32-bit unsigned integer
+    """
+    return crc32(data, init_value) & ChecksumConfig.CHECKSUM_XOR_OUT
```
LeStarch, line 63: This pattern arises again and again.  Should we put it somewhere reusable?  Perhaps as a future work item.

src/fprime_gds/common/dp/decoder.py
```diff
@@ -0,0 +1,293 @@
...
+            CRCError: If header checksum validation fails
+        """
+        header = get_dp_header_type()()
+        max_header_size = header.getMaxSize()
+        header_bin_data = file_handle.read(max_header_size)
+        header.deserialize(header_bin_data, 0)
+        
+        # Get actual header size and adjust file position
+        actual_header_size = header.getSize()
+        if actual_header_size < max_header_size:
```
LeStarch: So the header is variable size?

src/fprime_gds/common/dp/decoder.py
```diff
@@ -0,0 +1,293 @@
...
+            3. Get actual bytes consumed via getSize()
+            4. Seek to correct absolute position
+            
+            This avoids cumulative position errors from relative seeking.
+            """
+
+            element_instance = element_type()
+
+            # For types that may have variable length, read max size and adjust
+            if issubclass(element_type, (StringType, SerializableType, ArrayType)):
```
LeStarch: Why does it matter?  Each item supports max size ang get size, is the small boost in efficiency worth the extra branch to maintain?

src/fprime_gds/common/dp/decoder.py
```diff
@@ -0,0 +1,293 @@
...
+                file_handle.seek(start_pos + actual_size)
+            else:
+                # For fixed-size types, just read the exact size
+                element_data = file_handle.read(element_instance.getMaxSize())
+                element_instance.deserialize(element_data, 0)
+
+            return element_instance
+
+        # decode based on whether it's an array or scalar
+        if record_template.get_is_array():
```
LeStarch, line 179: Are arrays dynamic in length?  Otherwise I'd expect you to be able to just read the array as fixed size data.

src/fprime_gds/common/dp/validator.py
```diff
@@ -0,0 +1,245 @@
...
+        """Process and validate a data product file.
+        
+        Args:
+            data_product_path: Path to the data product file to validate
+            
+        Returns:
+            bool: True if validation succeeds, False otherwise
+        """
+        try:
+            dp_f = open(data_product_path, 'rb')
```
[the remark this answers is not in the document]
LeStarch: This definitely should be addressed.

src/fprime_gds/common/utils/cleanup.py
```diff
@@ -0,0 +1,20 @@
+
+def globals_cleanup():
+    """Cleans up all global/cached constructs.
+    
+    This is useful for example to start fresh between tests that may load
+    some of these up.
+    """
+    from fprime_gds.common.utils.config_manager import ConfigManager
```
LeStarch: Why used deferred importation here?

test/fprime_gds/common/dp/test_dp_data/dictionary.json
```diff
@@ -154,6 +154,11 @@
             "signed" : false
           },
           "enumeratedConstants" : [
+            {
+              "name" : "PROC_TYPE_HARD_ZERO",
```
LeStarch: This should not be needed any more please delete.

LeStarch (commented):
LGTM, some clarifying questions.

Also, why do we need a separate validator and decoder.  Couldn't validation be decoding w/o the write to file?

src/fprime_gds/common/dp/decoder.py
thomas-bc: yes, needs fixing

src/fprime_gds/common/dp/decoder.py
thomas-bc: Ok, will remove the if/else

src/fprime_gds/common/dp/decoder.py
thomas-bc: Yes, this is the array keyword in FPP, not an array type

https://nasa.github.io/fpp/fpp-users-guide.html#Defining-Components_Data-Products_Array-Records

LeStarch (approved):

PR #281, opened 2026-01-31
Add `uplink_sequence()` method to `IntegrationTestAPI`

nateinaction:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Hey team, This PR adds `uplink_sequence()` and `uplink_sequence_and_await_completion()` helper methods to the `IntegrationTestAPI` class. `uplink_sequence()` compiles text sequence files into bins and then passes them to the `uplink_file()` method.

## Rationale

I'm writing tests that verify sequences are cancelled under certain conditions. In our repo I have access to a `.seq` text file and I think this kind of helper feels more natural living upstream.

I have run tests using `uplink_sequence_and_await_completion()` and verified that they are uplinking and eventing successfully.

## Future Work

I didn't feel like there was a consistent method for catching and reraising errors in this class. Happy to take feed back on the implementation.

### DIFF
--- a/src/fprime_gds/common/testing_fw/api.py
+++ b/src/fprime_gds/common/testing_fw/api.py
@@ -13,6 +13,7 @@ import time
 from pathlib import Path
 import shutil
 import json
+import tempfile
 
 from fprime_gds.common.models.serialize.time_type import TimeType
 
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):
This is great, thanks a lot for contributing back. Was this tested end-to-end locally? It'd be good if we could get some testing in there for CI...

PR #282, opened 2026-01-31
Confetti 🎉 

nateinaction:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**| 🎊  |
|**_Documentation Included (y/n)_**| 🥳  |

---
## Change Description

On receiving the first packet, the GDS will now show confetti.

## Rationale

Who says we can't have fun right? And @Mikefly123 said I should...

https://github.com/user-attachments/assets/f7928eaa-4d33-4149-83a9-02529200bf4a

## Future Work

Merge it in before anyone cancels fun!


thomas-bc (commented):
Ha! No one would dare cancelling fun. Maybe if this were enabled only when using a `--confetti` option...... 👀  (or env var might be easier to retrieve in JS?)

nateinaction:
Haha! Thanks! I spent some more time looking into how we might want to pass a command line flag (or env var) into the vue components. It does not appear that the GDS currently supports modifying vue config at runtime including the modifications made recently in #267. Additionally based on how the PROVES team uses the GDS (automatic UART discovery), adding this option to the existing settings menu did not seem viable since we would likely see a first packet before navigating to the settings tab and changing the value. 

I played around with modifying the config at runtime and it's certainly possible but I think this is potentially an important feature and I would hate to build something simply to add a confetti addon and later find that the interface does not satisfy the requirements for a more valuable feature. As I moved toward productionalizing this I kept feeling regret for future maintainers who would bear the burden of ongoing maintenance and security on the confetti library. I don’t think the benefit here (beyond a fun easter-egg) justifies that cost.

I’ll close this PR for now, I'm happy to reopen or let someone else cross the finish line with it. I look forward to seeing you this week at the Advanced Workshop. 🖖

PR #284, opened 2026-02-12
Add automatic decoding for hashed files

youio:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
https://github.com/nasa/fprime/issues/4666
n
n

---
## Change Description

This PR adds a check for hashed files during `EventData` initialization and automatically decodes and replaces any that are found for display to the user. It also adds a new variable `BUILD_DIR` to the environment of the flask app to allow the new code to find the hashed files no matter where `fprime-gds` is called.

## Rationale

As outlined in issue [4666](https://github.com/nasa/fprime/issues/4666) on the fprime repository, this change allows users to benefit from using hashed file names without needing to manually call `hash-to-file` whenever an assert is triggered.

### DIFF
--- a/src/fprime_gds/common/data_types/event_data.py
+++ b/src/fprime_gds/common/data_types/event_data.py
@@ -80,16 +80,14 @@ class EventData(sys_data.SysData):
             except ValueError:
                 continue
 
-            hash_file = os.environ['HASH_FILE']
-            with open(hash_file) as file_handle:
-                lines = filter(
-                    lambda line: hash_value == int(line.split(" ")[-1], 0),
-                    file_handle.readlines(),
-                )
-            line = next(lines, None)
-            if line is not None:
-                file = Path(line.split(':')[0].strip()).absolute()
-                self.args[index].val = str(file)
+            hash_file = Path(os.environ['HASH_FILE'])
+            with hash_file.open() as file_handle:
+                for line in file_handle:
+                    if hash_value == int(line.split(" ")[-1], 0):
+                        rel_file_path = line.split(':')[0].strip()
+                        abs_file_path = (hash_file.parent.parent / rel_file_path).absolute()
+                        arg.val = str(abs_file_path) if abs_file_path.exists() else rel_file_path
+                        break
 
     def get_args(self):
         return self.args
--- a/src/fprime_gds/common/data_types/event_data.py
+++ b/src/fprime_gds/common/data_types/event_data.py
@@ -68,7 +68,7 @@ class EventData(sys_data.SysData):
             )
 
     def _decode_hashed_files(self):
-        "Searches event args for hashed files and replaces them with their corresponding file names"
+        "Searches event args for hashed files and replaces each with its corresponding file name"
 
         args_template = self.template.get_args()
         for index, arg in enumerate(self.args):
@@ -81,13 +81,13 @@ class EventData(sys_data.SysData):
                 continue
 
             hash_file = Path(os.environ['HASH_FILE'])
-            
             with open(hash_file) as file_handle:
                 lines = filter(
                     lambda line: hash_value == int(line.split(" ")[-1], 0),
                     file_handle.readlines(),
                 )
-            self.args[index].val = next(lines).split(':')[0].strip()
+            if len(list(lines)) > 0:
+                self.args[index].val = next(lines).split(':')[0].strip()
 
     def get_args(self):
         return self.args
--- a/src/fprime_gds/common/data_types/event_data.py
+++ b/src/fprime_gds/common/data_types/event_data.py
@@ -11,7 +11,10 @@ from fprime_gds.common.models.serialize import time_type
 
 from fprime_gds.common.data_types import sys_data
 from fprime_gds.common.utils.string_util import format_string_template
-import subprocess
+from fprime.fbuild.builder import Build
+from pathlib import Path
+
+import os
 
 class EventData(sys_data.SysData):
     """
@@ -46,21 +49,27 @@ class EventData(sys_data.SysData):
             return
 
         args_template = self.template.get_args()
-        for index, arg in enumerate(event_args):
+        if event_temp.name.startswith('AF_'):
+            for index, arg in enumerate(event_args):
 
-            # determine if args contain a hashed file
-            if args_template[index][0] != 'file':
-                continue
-            try:
-                int(arg.val, 16)
-            except ValueError:
-                continue
+                # determine if args contain a hashed file
+                if args_template[index][0] != 'file':
+                    continue
+                try:
+                    hash_value = int(arg.val, 16)
+                except ValueError:
+                    continue
 
-            # if so decode the filename and insert into arg
-            command = ['fprime-util', 'hash-to-file', arg.val]
-            result = subprocess.run(command, capture_output=True, text=True)
-            file = result.stdout.split('\n')[-2].split(':')[0].strip()
-            event_args[index].val = file
+                # if so decode the filename and insert into arg
+                build_dir = Path(os.environ['BUILD_DIR'])
+                parent_dir = build_dir.parent
+
+                build = Build(0, Build.find_nearest_parent_project(build_dir))
+                build.build_dir = build_dir
+                lines = build.find_hashed_file(hash_value)
+
+                file = parent_dir / Path(lines[0].split(':')[0].strip())
+                event_args[index].val = str(file)
 
         if event_temp.format_str == "":
             self.display_text = str(
--- a/src/fprime_gds/common/data_types/event_data.py
+++ b/src/fprime_gds/common/data_types/event_data.py
@@ -11,7 +11,6 @@ from fprime_gds.common.models.serialize import time_type
 
 from fprime_gds.common.data_types import sys_data
 from fprime_gds.common.utils.string_util import format_string_template
-from pathlib import Path
 
 import os
 
@@ -83,7 +82,7 @@ class EventData(sys_data.SysData):
                 continue
 
             hash_file = os.environ['FPRIME_HASHES_TXT_FILE']
-            with hash_file.open() as file_handle:
+            with open(hash_file) as file_handle:
                 for line in file_handle:
                     if hash_value == int(line.split(" ")[-1], 0):
                         arg.val = line.split(':')[0].strip()
--- a/src/fprime_gds/common/data_types/event_data.py
+++ b/src/fprime_gds/common/data_types/event_data.py
@@ -11,7 +11,6 @@ from fprime_gds.common.models.serialize import time_type
 
 from fprime_gds.common.data_types import sys_data
 from fprime_gds.common.utils.string_util import format_string_template
-from fprime.fbuild.builder import Build
 from pathlib import Path
 
 import os
@@ -48,7 +47,10 @@ class EventData(sys_data.SysData):
             self.display_text = event_temp.description
             return
 
-        if event_temp.name.startswith('AF_ASSERT') or event_temp.name == "AF_UNEXPECTED_ASSERT":
+        if (
+            (event_temp.name.startswith('AF_ASSERT') or event_temp.name == "AF_UNEXPECTED_ASSERT") and
+            'HASH_FILE' in os.environ
+        ):
             self._decode_hashed_files()
                 
         if event_temp.format_str == "":
@@ -66,11 +68,11 @@ class EventData(sys_data.SysData):
             )
 
     def _decode_hashed_files(self):
-        args_template = self.template.get_args()
+        "Searches event args for hashed files and replaces them with their corresponding file names"
 
+        args_template = self.template.get_args()
         for index, arg in enumerate(self.args):
 
-            # determine if args contain a hashed file
             if args_template[index][0] != 'file':
                 continue
             try:
@@ -78,16 +80,14 @@ class EventData(sys_data.SysData):
             except ValueError:
                 continue
 
-            # if so decode the filename and insert into arg
-            build_dir = Path(os.environ['BUILD_DIR'])
-            parent_dir = build_dir.parent
-
-            build = Build(0, Build.find_nearest_parent_project(build_dir))
-            build.build_dir = build_dir
-            lines = build.find_hashed_file(hash_value)
-
-            file = parent_dir / Path(lines[0].split(':')[0].strip())
-            self.args[index].val = str(file)
+            hash_file = Path(os.environ['HASH_FILE'])
+            
+            with open(hash_file) as file_handle:
+                lines = filter(
+                    lambda line: hash_value == int(line.split(" ")[-1], 0),
+                    file_handle.readlines(),
+                )
+            self.args[index].val = next(lines).split(':')[0].strip()
 
     def get_args(self):
         return self.args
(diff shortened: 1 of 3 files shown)


src/fprime_gds/common/data_types/event_data.py
```diff
@@ -40,10 +40,29 @@ def __init__(self, event_args, event_time, event_temp):
...
-        elif event_temp.format_str == "":
-            args_template = self.template.get_args()
+            return
+
+        args_template = self.template.get_args()
+        for index, arg in enumerate(event_args):
+
+            # determine if args contain a hashed file
+            if args_template[index][0] != 'file':
+                continue
```
celskeggs: Are we sure we don't want to restrict this to only apply to assertion events, rather than any event with a "file" argument?

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -40,10 +40,29 @@ def __init__(self, event_args, event_time, event_temp):
...
+            try:
+                int(arg.val, 16)
+            except ValueError:
+                continue
+
+            # if so decode the filename and insert into arg
+            command = ['fprime-util', 'hash-to-file', arg.val]
```
celskeggs: The implementation of `hash-to-file` is very simple. See here:

https://github.com/nasa/fprime-tools/blob/864dcb2575d861687fc7bbf8b6d454bf657ed885/src/fprime/fbuild/builder.py#L181-L193

Is it worth shelling out like this?

EDIT: Apologies, I missed the note about this in the Future Work section. But I do think it would be worth seriously considering using the `find_hashed_file` function.

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -40,10 +40,29 @@ def __init__(self, event_args, event_time, event_temp):
...
+            result = subprocess.run(command, capture_output=True, text=True)
```
celskeggs: What if `fprime-gds` is executed outside of the project directory? Would this correctly resolve the build directory?

youio:
@celskeggs Good points, I pushed some changes that I think should fix all of those issues. Thanks!

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -40,10 +43,35 @@ def __init__(self, event_args, event_time, event_temp):
...
         self.template = event_temp
+
         if event_args is None:
             self.display_text = event_temp.description
-        elif event_temp.format_str == "":
-            args_template = self.template.get_args()
+            return
+
+        args_template = self.template.get_args()
+        if event_temp.name.startswith('AF_'):
```
celskeggs: Are we sure that `AF_` is a sufficient check to make sure an event is really an assert? I would consider checking for `AF_ASSERT_*` or `AF_UNEXPECTED_ASSERT`.

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -40,10 +43,35 @@ def __init__(self, event_args, event_time, event_temp):
...
+                    hash_value = int(arg.val, 16)
+                except ValueError:
+                    continue
+
+                # if so decode the filename and insert into arg
+                build_dir = Path(os.environ['BUILD_DIR'])
+                parent_dir = build_dir.parent
+
+                build = Build(0, Build.find_nearest_parent_project(build_dir))
+                build.build_dir = build_dir
```
celskeggs: I might not understand the interface to builder.Build, but this doesn't seem quite right to me. Typically `build_dir` is set by the `Build` class via `load` or `invent`. While this might work, I'm worried that it may not be a robust long-term solution, since changes to the internal representation of `Build` could break it.

In addition, using `load` or `invent` might be the wrong idea too, since those expect a full build cache to be available, and someone might be running the F Prime GDS with just a copy of the dictionary, not a copy of the build cache.

Maybe it would be better to refactor the `find_hashed_file` function so that there's a common version that can be used both by `Build` and by this code. I'm not totally sure, though...

src/fprime_gds/executables/run_deployment.py
```diff
@@ -122,6 +122,7 @@ def launch_html(parsed_args):
             "FLASK_APP": "fprime_gds.flask.app",
             "STANDARD_PIPELINE_ARGUMENTS": "|".join(reproduced_arguments),
             "SERVE_LOGS": "YES",
+            "BUILD_DIR": parsed_args.deployment.parent.parent
```
celskeggs: Isn't it usually optional to pass the deployment to fprime-gds? What happens if the user doesn't?

celskeggs (commented):
Thanks for the changes! I do think this is better, but I'm not sure it's totally robust yet. We may want the F Prime core team to weigh in on how we should handle some of these things.

src/fprime_gds/executables/run_deployment.py
thomas-bc: Yes it is optional. We probably need to add a `--hash-file` flag or something, for users to provide it in the case that it can't be detected.

To clarify, `parsed_args.deployment` does get auto-populated if it can be detected (i.e. if a valid `build-artifact` dir is found). But in our case, we still want this feature to be available even if ran with just a `Dictionary.json` and `hashes.txt` file.

src/fprime_gds/common/data_types/event_data.py
thomas-bc: > Maybe it would be better to refactor the find_hashed_file function so that there's a common version that can be used both by Build and by this code

I like this! If we have a `--hash-file` providing us with the file (see other comment), we would not need to depend on `Build` at all. Given that the commonality is really just a [filter function](https://github.com/thomas-bc/fprime-tools/blob/b16a59e4e7721ab243b6f3e8d029978de306e341/src/fprime/fbuild/builder.py#L190), we could also simply just redefine the filter and drop the dependency on `fprime-tools` altogether.

thomas-bc (changes requested):
This is super helpful, thanks a lot! A few comments for your consideration. As @celskeggs is hinting at, I think we'll need a way for users to provide a hash file themselves and not necessarily rely on the `build-artifacts` detection. From there, we can re-organize a few things to make this pretty self-contained.

thomas-bc:
> As @celskeggs is hinting at, I think we'll need a way for users to provide a hash file themselves and not necessarily rely on the build-artifacts detection.

@celskeggs as the requestor of that feature, do you agree here? Since both `dict.json` and `hashes.txt` are in the `build-artifacts`, an alternative is to copy/require the entire `build-artifacts` when using that feature. Thoughts?

celskeggs:
@thomas-bc In my context, I can definitely provide a full build-artifacts directory, or at least one that contains hashes.txt and the dictionary in the right places. No complaints about the idea of `--hash-file`, though. But I cannot rely on providing an entire FSW build directory, just the build-artifacts.

timcanham:
It's typical (and sometimes required) to deliver the flight dictionaries without the source/build tree, so there shouldn't be a dependency to decode the hashes.

youio:
Thanks everyone for the suggestions and feedback. @thomas-bc I implemented the --hash-file flag and integrated all of your other suggestions. The code still tries to detect the hash file automatically if not provided and will simply output the original hash code if it fails. Let me know if you find any other issues with the implementation or if there's anything else I should add. Thanks!

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -55,6 +67,30 @@ def __init__(self, event_args, event_time, event_temp):
...
+            hash_file = os.environ['HASH_FILE']
+            with open(hash_file) as file_handle:
+                lines = filter(
+                    lambda line: hash_value == int(line.split(" ")[-1], 0),
+                    file_handle.readlines(),
+                )
+            line = next(lines, None)
+            if line is not None:
+                file = Path(line.split(':')[0].strip()).absolute()
+                self.args[index].val = str(file)
```
celskeggs: Suggestion... not tested, but might be a simpler way to do this?

```
            hash_file = os.environ['HASH_FILE']
            with open(hash_file) as file_handle:
                for line in file_handle:
                    if hash_value == int(line.split(" ")[-1], 0):
                        file = Path(line.split(':')[0].strip()).absolute()
                        arg.val = str(file)
                        break
```

celskeggs: Or... also, wouldn't the path end up being resolved relative to the CWD of F Prime GDS? That might result in completely bogus paths. Maybe you want:

```
            hash_file = Path(os.environ['HASH_FILE'])
            with hash_file.open() as file_handle:
                for line in file_handle:
                    if hash_value == int(line.split(" ")[-1], 0):
                        file_name = line.split(':')[0].strip()
                        file = hash_file.parent.parent / file_name
                        arg.val = str(file.absolute()) if file.exists() else file_name
                        break
```

(Also not tested, just for consideration)

youio: True, the filter was from Build.find_hashed_file() but this makes a lot more sense here. Also good point with the .absolute() call. I went ahead and incorporated these suggestions and everything appears to be working properly.

celskeggs (approved):
Looks good to me!

src/fprime_gds/executables/cli.py
```diff
@@ -478,6 +478,31 @@ def handle_arguments(self, args, **kwargs):
         args.deployment = child_directories[0]
         return args
 
+class HashFileParser(ParserBase):
+    """Parser for detecting and loading the hashes.txt file for hash decoding"""
+
```
thomas-bc:
```suggestion

    DESCRIPTION = "Hash file options"

```

src/fprime_gds/executables/cli.py
```diff
@@ -478,6 +478,31 @@ def handle_arguments(self, args, **kwargs):
...
+    def get_arguments(self)-> Dict[Tuple[str, ...], Dict[str, Any]]:
+        return {
+            ("--hash-file",): {
+                "dest": "hash_file",
+                "action": "store",
+                "required": False,
+                "type": str,
+                "help": "File containing map between hash codes and their corresponding program files",
```
thomas-bc:
```suggestion
                "help": "Path to hashes.txt file map (found under build-artifacts dir by default)",
```

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -40,10 +42,20 @@ def __init__(self, event_args, event_time, event_temp):
...
         self.template = event_temp
+
         if event_args is None:
             self.display_text = event_temp.description
-        elif event_temp.format_str == "":
+            return
+
+        if (
+            (event_temp.name.startswith('AF_ASSERT') or event_temp.name == "AF_UNEXPECTED_ASSERT") and
+            'HASH_FILE' in os.environ
```
thomas-bc: Let's rename this env var to `FPRIME_HASHES_TXT_FILE` so that it's easier to understand if not looking at the code from this context.

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -55,6 +67,28 @@ def __init__(self, event_args, event_time, event_temp):
                 event_temp.format_str, tuple([arg.val for arg in event_args])
             )
 
+    def _decode_hashed_files(self):
+        "Searches event args for hashed files and replaces each with its corresponding file name"
+
+        args_template = self.template.get_args()
+        for index, arg in enumerate(self.args):
+
+            if args_template[index][0] != 'file':
```
thomas-bc, line 77: For the sake of being defensive I'd recommend we check `args_template[index]` is not empty before accessing it. Probably `continue` if it is empty? Unlikely in the current setting, but let's be future-proof! 😃

src/fprime_gds/common/data_types/event_data.py
```diff
@@ -55,6 +67,28 @@ def __init__(self, event_args, event_time, event_temp):
...
+                hash_value = int(arg.val, 16)
+            except ValueError:
+                continue
+
+            hash_file = Path(os.environ['HASH_FILE'])
+            with hash_file.open() as file_handle:
+                for line in file_handle:
+                    if hash_value == int(line.split(" ")[-1], 0):
+                        rel_file_path = line.split(':')[0].strip()
+                        abs_file_path = (hash_file.parent.parent / rel_file_path).absolute()
```
thomas-bc: Testing locally, I never get `abs_file_path` to work. Does it for you? Could be that I'm building in `fprime/Ref` which has a different project layout.

src/fprime_gds/executables/cli.py
```diff
@@ -478,6 +478,31 @@ def handle_arguments(self, args, **kwargs):
...
+        }
+    
+    def handle_arguments(self, args, **kwargs):
+        if args.hash_file:
+            args.hash_file = Path(args.hash_file)
+            if not args.hash_file.exists():
+                msg = f"[ERROR] hash file location {args.hash_file} does not exist"
+                print(msg, file=sys.stderr)
+                sys.exit(-1)
+        elif args.deployment:
```
thomas-bc: One recommendation: you can look at how the `DictionaryParser` does it and borrow the logic to be a bit more robust - it inherits from `DetectionParser` which finds the deployment, and uses that (with the `super()` call) to do its thing based on finding the deployment. Here, you're checking `args.deployment` which may or may not have been filled (it is dependent on the order in which parsers are listed in `CONSTITUENTS`, which is I assume why you've had to list HashFileParser after DictionaryParser).

src/fprime_gds/executables/cli.py
```diff
@@ -1148,6 +1174,7 @@ class CommParser(CompositeParser):
 
     CONSTITUENTS = [
         DictionaryParser,  # needed to get types from dictionary for framing
+        HashFileParser,
```
thomas-bc: Does the Comm layer need this? I don't think so since it shouldn't need to decode events, but I may be wrong.

src/fprime_gds/executables/cli.py
thomas-bc: Happy to merge this PR without the above recommendation; but figured I'd call it out if you're interested.

thomas-bc (changes requested):
Last round of nits and recommendations - I'm expecting to be ready to merge once we've addressed them. And thanks again!

src/fprime_gds/common/data_types/event_data.py
thomas-bc: I might be in favor of taking out abs_file_path altogether:
- can lead to inconsistent user experience if file paths come relative sometimes, absolute others
- absolute can be hard to read if they're too long. relative is easier to identify, and it's what we have by default

But that's also likely just personal preference. I'll let you make that call!

thomas-bc:
@youio if you don't have time to get through this, I am happy to help

src/fprime_gds/executables/cli.py
```diff
@@ -478,6 +478,33 @@ def handle_arguments(self, args, **kwargs):
         args.deployment = child_directories[0]
         return args
 
+class HashFileParser(ParserBase):
```
thomas-bc:
```suggestion
class HashFileParser(DetectionParser):
```

youio:
@thomas-bc Hi apologies I just incorporated your requests. I ended up removing the absolute path logic like you said. I thought it might be convenient but its much trickier to implement robustly than I thought and I think the relative paths are sufficient. For the `args.deployment` logic I realized that calling the `DetectionParser` `handle_arguments()` in the `HashFileParser` directly might throw the "deployment not found" error if the user is only using the dictionary and isn't using hashed files, so I made it inherit from `DictionaryParser` to check if the deployment is necessary before searching for it. Let me know if that solution works or if any last changes are needed. Thanks!

thomas-bc (approved):
Looks good to me! Thanks a lot!

PR #286, opened 2026-02-26
Update fw_testing pytest integration to include using default configuration file

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**|  https://github.com/nasa/fprime/issues/4770 |
|**_Has Unit Tests (y/n)_**|  No |
|**_Documentation Included (y/n)_**| y (code comments) |

---
## Change Description

1. Updated `fprime_test_api_session` in `pytest_integration.py` to include default GDS configuration using the ConfigDrivenParser  class 
2.  Updated ConfigDrivenParser class to check the `FPRIME_GDS_CONFIG_PATH` as part of its default behavior (previously the default file path was included in the class implementation but the environment variable override of that path was only checked in run_deployment.py)

## Rationale

The rationale for updating `fprime_test_api_session`  was so that running `pytest` would pick up the same GDS defaults as `fprime-gds`  - without that feature in my environment, we'd need to have a bunch of different pytest configuration files to ensure the correct dictionary got used for different deployments being tested 

Updating the ConfigDrivenParser class was required to avoid environment variable name that could override the default configuration path being specified and handled in multiple locations (previously only in run_deployment.py, but the updates to pytest_integration.py would've also required it) 

To test: 
1. Create a fprime-gds.yml with unique configuration in a directory with tests, run pytest and verify those settings are picked up 
2. Create a fprime-gds.yml in a different location, set the FPRIME_GDS_CONFIG_PATH environment variable with the path to that file,  run pytest and verify those settings are picked up 
3. Verify pytest uses normal defaults when FPRIME_GDS_CONFIG_PATH  is not set and there's no prime-gds.yml in the current directory 

## Future Work

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -18,6 +18,7 @@ import os
 import platform
 import re
 import sys
+import pathlib
 
 import yaml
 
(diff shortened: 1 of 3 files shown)


src/fprime_gds/executables/cli.py
```diff
@@ -1038,10 +1078,21 @@ def handle_arguments(self, args, **kwargs):
...
         )
+        config = ConfigManager.get_instance()
+        # Update config to use type definitions defined in the JSON dictionary
+        if dictionaries.typedefs_name:
+            for type_name, type_dict in dictionaries.typedefs_name.items():
+                config.set_type(type_name, type_dict)
+        if dictionaries.constant_name:
+            for name, value in dictionaries.constant_name.items():
+                config.set_constant(name, value)
+        args.dictionaries = dictionaries
```
thomas-bc, line 1095: This blob of change looks like a recent change has been reverted - you likely branched off before it was made. Could we revert it back to the current devel state?

thomas-bc (changes requested):

LeStarch:
Closing as stale.

PR #289, opened 2026-03-04
Optimize DataProductDecoder for large numerical arrays

olabie2:
| | |
|:---|:---|
|**_Related Issue(s)_**| [nasa/fprime#4784](https://github.com/nasa/fprime/issues/4784) |
|**_Has Unit Tests (y/n)_**| y (existing tests cover this change — 25/25 pass) |

---
## Change Description

Add a batch deserialization fast path in DataProductDecoder.decode_record() for arrays of NumericalType (U8, U16, U32, U64, I8, I16, I32, I64, F32, F64).

Instead of calling read_element() once per array element, the entire array is read in a single I/O operation and deserialized with one struct.unpack_from() call using a batched format string via a dedicated helper method [_batch_deserialize_numerical_array()](file:///home/uolabie/open-source/forked/fprime-gds/src/fprime_gds/common/dp/decoder.py#106-133). Non numerical types (structs, enums, strings) continue to use the original per element path.

The JSON output format is unchanged.

## Rationale

Fixes a performance bottleneck where decoding a 12MB data product with U8 arrays took ~45 seconds due to [read_element()](file:///home/uolabie/open-source/forked/fprime-gds/src/fprime_gds/common/dp/decoder.py#190-207) being called ~11.5 million times. Benchmarked by @kubiak-jpl at 1.451s after this change (~31x speedup).

## Benchmark

Profiling the U8 array test file (6 records × 300 U8 elements) decoded 200 times:

```
         317001 function calls (309801 primitive calls) in 0.241 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      200    0.010    0.000    0.241    0.001 decoder.py:235(decode)
     1200    0.009    0.000    0.102    0.000 decoder.py:160(decode_record)
      200    0.002    0.000    0.095    0.000 decoder.py:134(decode_header)
     1200    0.054    0.000    0.060    0.000 decoder.py:106(_batch_deserialize_numerical_array)
    15600    0.008    0.000    0.025    0.000 type_base.py:98(to_jsonable)
    20800    0.013    0.000    0.015    0.000 numerical_types.py:35(getSize)
    10800    0.007    0.000    0.011    0.000 numerical_types.py:57(deserialize)
    16000    0.006    0.000    0.010    0.000 type_base.py:45(__repr__)
```

[read_element](file:///home/uolabie/open-source/forked/fprime-gds/src/fprime_gds/common/dp/decoder.py#190-207) no longer appears in the profile — numerical arrays are fully handled by [_batch_deserialize_numerical_array](file:///home/uolabie/open-source/forked/fprime-gds/src/fprime_gds/common/dp/decoder.py#106-133) in a single I/O + struct.unpack_from call per record.

@kubiak-jpl also independently benchmarked this PR against the original 12MB data product from [#4784](https://github.com/nasa/fprime/issues/4784) and confirmed the improvement (45.6s → 1.45s).

Run the decoder test suite:

`pytest test/fprime_gds/common/dp/test_decoder.py -v`

- All 25 existing tests pass without modification, including array type tests (test_decode_u8_array, test_decode_u32_array, test_decode_data_array, test_decode_fpp_array).
- The `issubclass(record_type, NumericalType)` check ensures only fixed-size types use the fast path. Variable-length or complex types fall through to the original loop.
- Review that the batch format string construction correctly strips and re-applies the big-endian prefix.

## Future Work

- The JSON output currently repeats `{"value": N, "type": "U8"}` for every element. A future optimization could store the type name once per array record rather than per element, further reducing memory and serialization overhead (as noted in the original issue discussion).

- An [issue](https://github.com/nasa/fprime/issues/4806) is already open for this.

### DIFF
--- a/src/fprime_gds/common/dp/decoder.py
+++ b/src/fprime_gds/common/dp/decoder.py
@@ -14,11 +14,12 @@ Key differences from the original implementation:
 @date: January 2026
 """
 
+import dataclasses
 import json
+import struct
 import sys
 from pathlib import Path
 from typing import Dict, List, Any, Optional
-import dataclasses
 
 from fprime_gds.common.dp.common import (
     ChecksumConfig,
@@ -26,6 +27,7 @@ from fprime_gds.common.dp.common import (
     get_dp_header_type,
 )
 from fprime_gds.common.models.dictionaries import Dictionaries
+from fprime_gds.common.models.serialize.numerical_types import NumericalType
 from fprime_gds.common.utils.config_manager import ConfigManager
 from fprime_gds.common.templates.dp_record_template import DpRecordTemplate
 
@@ -101,6 +103,34 @@ class DataProductDecoder:
         else:
             self.output_json_path = output_json_path
         
+    @staticmethod
+    def _batch_deserialize_numerical_array(file_handle, record_type, array_size) -> List[Dict[str, Any]]:
+        """Batch-read and deserialize an array of fixed-size numerical types.
+
+        Reads the entire array in one I/O operation and uses struct.unpack_from
+        with a batched format string, avoiding per-element overhead.
+
+        Args:
+            file_handle: file handle positioned at the start of the array data
+            record_type: NumericalType subclass (e.g. U8Type, U32Type, F64Type)
+            array_size: number of elements in the array
+
+        Returns:
+            List of JSON-serializable dicts, one per element
+        """
+        element_size = record_type.getSize()
+        total_bytes = array_size * element_size
+        buffer = file_handle.read(total_bytes)
+
+        # Build a batch format string (e.g. ">10000B" for 10000 U8s)
+        base_format = record_type.get_serialize_format().lstrip(">")
+        batch_format = f">{array_size}{base_format}"
+        values = struct.unpack_from(batch_format, buffer, 0)
+
+        # Build JSON list with type name computed once
+        type_name = record_type.get_canonical_name()
+        return [{"value": val, "type": type_name} for val in values]
+
     def decode_header(self, file_handle) -> Dict[str, Any]:
         """Decode the data product header.
 
@@ -184,12 +214,17 @@ class DataProductDecoder:
             array_size = array_size_type.val
 
             record['Size'] = array_size
-            record['Data'] = []
 
-            # Read each array element
-            for _ in range(array_size):
-                element_instance = read_element(record_type)
-                record['Data'].append(element_instance.to_jsonable())
+            if issubclass(record_type, NumericalType) and array_size > 0:
+                record['Data'] = self._batch_deserialize_numerical_array(
+                    file_handle, record_type, array_size
+                )
+            else:
+                # General path for complex or variable-length types
+                record['Data'] = []
+                for _ in range(array_size):
+                    element_instance = read_element(record_type)
+                    record['Data'].append(element_instance.to_jsonable())
         else:
             # For scalar records, read the single value
             element_instance = read_element(record_type)
@@ -267,4 +302,3 @@ class DataProductDecoder:
         except Exception as e:
             print(f"Unexpected error: {e}", file=sys.stderr)
             raise
-


kubiak-jpl:
I like it. I do think it's worth considering dropping the per element type information though. My tests showed that generating that structure for every byte of the file was a significant amount of processing and I struggle to think of a reason why it needs to be repeated for each sample

src/fprime_gds/common/dp/decoder.py
```diff
@@ -184,12 +186,31 @@ def read_element(element_type):
             array_size = array_size_type.val
 
             record['Size'] = array_size
-            record['Data'] = []
 
-            # Read each array element
-            for _ in range(array_size):
-                element_instance = read_element(record_type)
-                record['Data'].append(element_instance.to_jsonable())
+            if issubclass(record_type, NumericalType) and array_size > 0:
```
thomas-bc, line 218: Could we pull this out into a helper method to make the code easier to follow? Makes it easier to maintain for us.

```python
if issubclass(record_type, NumericalType) and array_size > 0:
    record['Data'] = optimized_array_unpacking(<args>) # feel free to find a better name
```

src/fprime_gds/common/dp/decoder.py
thomas-bc, line 1: This looks awesome! Would you able to benchmark this improvement with the same method that @kubiak-jpl showed in https://github.com/nasa/fprime/issues/4784 ?

thomas-bc (changes requested):
LGTM! Just one quick nit about structure

kubiak-jpl:
@thomas-bc Any thoughts on the output JSON structure?

thomas-bc:
Yes happy to remove the repeated type info. Do you have a suggestion on how it should be structured instead? Maybe just a raw list of values instead of list of dicts? We just need to make sure the type info is somewhere.

kubiak-jpl:
Here's the profiling results I get running my script over the same product
```
(venv) kubiak@debian-13-base:~/machines/moonfall/home/source/fprime-voxl2-cams/scripts$ python3 -m cProfile ./fprime-image-dp.py -d ../build-artifacts/voxl2/Voxl2Cams/dict/Voxl2CamsTopologyDictionary.json -b ../Dp_268455936_17722
17218_00198277.fdp                                                                                                                                                                                                                   
         545155 function calls (427792 primitive calls) in 1.451 seconds                                                                                                                                                             
                                                                                                                                                                                                                                     
   Ordered by: cumulative time                                                                                                                                                                                                       
                                                                                                                                                                                                                                     
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)                                                                                                                                                              
    184/1    0.001    0.000    1.451    1.451 {built-in method builtins.exec}                                                                                                                                                        
        1    0.227    0.227    1.451    1.451 fprime-image-dp.py:1(<module>)                                                                                                                                                         
        1    0.001    0.001    1.205    1.205 fprime-image-dp.py:15(main)                                                                                                                                                            
        1    0.008    0.008    0.695    0.695 decoder.py:221(decode)                                                                                                                                                                 
      120    0.661    0.006    0.683    0.006 decoder.py:132(decode_record)                                                                                                                                                          
       29    0.000    0.000    0.198    0.007 __init__.py:1(<module>)                                                                                                                                                                
       30    0.188    0.006    0.189    0.006 fprime_image.py:61(from_json) 
```

Which is much better. However, if I replace the 

```diff
diff --git a/src/fprime_gds/common/dp/decoder.py b/src/fprime_gds/common/dp/decoder.py
index 6925891..2eeec89 100644
--- a/src/fprime_gds/common/dp/decoder.py
+++ b/src/fprime_gds/common/dp/decoder.py
@@ -202,9 +202,10 @@ class DataProductDecoder:
 
                 # Build JSON list with type name computed once
                 type_name = record_type.get_canonical_name()
-                record['Data'] = [
-                    {"value": val, "type": type_name} for val in values
-                ]
+                record['Data'] = {
+                    "type": type_name,
+                    "value": [val for val in values]
+                }
             else:
                 # General path for complex or variable-length types
                 record['Data'] = []
```

This translates to the following in the generated json file.
```
      "Data": {
        "type": "U8",
        "value": [
          11,
          11,
          12,
...
          0,             
          0,                
          0                 
        ]            
      }           
    },
```


This shaves an additional second off of the runtime and preserves the type information in the JSON. Part if this is not needing to create the dictionary objects for each element, and I am guessing part of it is not needing to unpack these dictionary object later in my processing code.

```
         545191 function calls (427828 primitive calls) in 0.473 seconds                                        
                                                                                                                  
   Ordered by: cumulative time                                                                                    
                                                                                                                  
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)                              
    184/1    0.001    0.000    0.473    0.473 {built-in method builtins.exec}                           
        1    0.008    0.008    0.473    0.473 fprime-image-dp.py:1(<module>)                            
        1    0.001    0.001    0.436    0.436 fprime-image-dp.py:15(main)                           
       29    0.001    0.000    0.207    0.007 __init__.py:1(<module>)                        
       30    0.176    0.006    0.176    0.006 {imwrite}                                                           
        1    0.008    0.008    0.112    0.112 decoder.py:222(decode)                      
      120    0.063    0.001    0.099    0.001 decoder.py:132(decode_record)                   
    228/6    0.001    0.000    0.089    0.015 <frozen importlib._bootstrap>:1349(_find_and_load)          
    228/6    0.000    0.000    0.089    0.015 <frozen importlib._bootstrap>:1304(_find_and_load_unlocked)
   216/10    0.000    0.000    0.088    0.009 <frozen importlib._bootstrap>:911(_load_unlocked)  
   178/10    0.000    0.000    0.087    0.009 <frozen importlib._bootstrap_external>:1020(exec_module)                                                                                                                               
   498/15    0.000    0.000    0.087    0.006 <frozen importlib._bootstrap>:480(_call_with_frames_removed) 
   414/23    0.000    0.000    0.057    0.002 {built-in method builtins.__import__}        
   178/50    0.000    0.000    0.054    0.001 <frozen importlib._bootstrap>:1390(_handle_fromlist)           
       30    0.007    0.000    0.043    0.001 __init__.py:120(dump)
38460/38100    0.005    0.000    0.033    0.000 encoder.py:415(_iterencode)
148500/38100    0.021    0.000    0.028    0.000 encoder.py:334(_iterencode_dict)
        1    0.000    0.000    0.025    0.025 __config__.py:1(<module>)
        1    0.000    0.000    0.023    0.023 dictionaries.py:1(<module>)
        1    0.001    0.001    0.020    0.020 dictionaries.py:168(load_dictionaries_into_config)
        1    0.000    0.000    0.019    0.019 dictionaries.py:64(load_dictionaries)
       90    0.000    0.000    0.018    0.000 decoder.py:162(read_element)
    31260    0.004    0.000    0.018    0.000 encoder.py:278(_iterencode_list)
  662/332    0.015    0.000    0.016    0.000 {built-in method builtins.sum}
        1    0.000    0.000    0.016    0.016 _arraypad_impl.py:1(<module>)
        1    0.000    0.000    0.016    0.016 _index_tricks_impl.py:1(<module>)
       30    0.015    0.001    0.016    0.001 fprime_image.py:61(from_json)
  420/240    0.000    0.000    0.016    0.000 serializable_type.py:150(getSize)
1680/1020    0.000    0.000    0.015    0.000 serializable_type.py:152(<genexpr>)
        1    0.000    0.000    0.015    0.015 ch_xml_loader.py:1(<module>)
      178    0.001    0.000    0.015    0.000 <frozen importlib._bootstrap_external>:1093(get_code)
        1    0.000    0.000    0.014    0.014 defmatrix.py:1(<module>)
      854    0.014    0.000    0.014    0.000 {built-in method _struct.unpack_from}
```

If we do make this change, we should also change the `UserData` field in the header to match the updated structure. With the diff above it still is  processed element by element
```
    "UserData": {
      "format": "{}",
      "description": "User-configurable data",
      "name": "UserData",
      "type": "UserData",
      "size": 32,
      "values": [
        {
          "value": 0,
          "type": "U8"
        },
        {
          "value": 0,
          "type": "U8"
        },
```

thomas-bc:
I like this structure! @olabie2 would you want to implement this?

thomas-bc:
With regards to the `UserData` field: There is a difference between [array records](https://nasa.github.io/fpp/fpp-users-guide.html#Defining-Components_Data-Products_Array-Records) (which is what we're optimizing here) and records that have a type which is an FPP array.
```
product record DataArrayRecord: U8 array
```
vs.
```
array MyU8ArrayType = [100] U8
product record DataArrayRecord: MyU8ArrayType
```
I believe the second approach (since it still relies on the base `to_jsonable()`) would still be decoded as such:
```
"values": [
        {
          "value": 0,
          "type": "U8"
        },
        {
          "value": 0,
          "type": "U8"
        },
```

I also believe UserData falls closer to the second approach, since it has a fixed configurable size. In this case, we may want to keep it that way and not optimize there. Can you confirm?

olabie2:
> I like this structure! @olabie2 would you want to implement this?

sure

olabie2:
> I also believe UserData falls closer to the second approach, since it has a fixed configurable size. In this case, we may want to keep it that way and not optimize there. Can you confirm?

Yeah, I think that’s right. The optimization only applies to array records (where get_is_array() is True and the element type is a NumericalType subclass). FPP array types like MyU8ArrayType use ArrayType, not NumericalType, so they don’t hit this path. UserData is part of the header and gets processed by decode_header() via to_jsonable(), so this change doesn’t affect it at all.

kubiak-jpl:
I have two concerns then
1. If I have a record of a serializable large member arrays then they won't get this optimization, are likely going to hit a similar performance wall as the one that triggered this PR
2. The format for array records vs. structures with arrays are different in both the json and the internal dictionary structure that the Data Product decoder provides

I think 2 would cause quite a bit of confusion and should be avoided. So maybe this PR should only cover the initial performance issue and the format change can go in another issue where everything is changed at once?

olabie2:
I second this, let's see what @thomas-bc think

thomas-bc:
Sounds good to me! Thanks to you both.

olabie2:
@thomas-bc You can now review it ,  it's optimized using the first method leaving the old json structure unchanged.

src/fprime_gds/common/dp/decoder.py
```diff
@@ -101,6 +103,34 @@ def __init__(self, dictionaries: Dictionaries, binary_file_path: str, output_jso
...
+        buffer = file_handle.read(total_bytes)
+
+        # Build a batch format string (e.g. ">10000B" for 10000 U8s)
+        base_format = record_type.get_serialize_format().lstrip(">")
+        batch_format = f">{array_size}{base_format}"
+        values = struct.unpack_from(batch_format, buffer, 0)
+
+        # Build JSON list with type name computed once
+        type_name = record_type.get_canonical_name()
+        return [{"value": val, "type": type_name} for val in values]
```
thomas-bc, line 132: One concern I have is with this line here, where we're re-encoding the `to_jsonable` format for an array, but in a completely separate file. I'll wait for the CCB to decide what to do with https://github.com/nasa/fprime/issues/4806 to know what to do about this. Sorry for the delay.

thomas-bc: A workaround could be that `_batch_deserialize` returns a constructed `ArrayType`, and we just call `to_jsonable` on its return value instead of constructing the list ourselves.

olabie2: Not a problem at all! That makes sense.
I’m happy to adjust the implementation afterward if needed.

thomas-bc:
Ok, sorry for the delay to you both. We have a lot going on.

#292 was merged so we should be ready to go forward with this PR here

There's a conflict now. @olabie2 would you be able to resolve it?

olabie2:
Sure I will do it

olabie2:
Hey @thomas-bc  after reviewing #292 (merged), I can confirm it fully incorporates the optimization from this PR. 


What #292 covers
The batch struct.unpack_from deserialization that was in my _batch_deserialize_numerical_array() is now built directly into ArrayType.deserialize() ([array_type.py](https://github.com/nasa/fprime-gds/blob/devel/src/fprime_gds/common/models/serialize/array_type.py#L149-L179))
The decoder now delegates array records to ArrayType.construct_type() instead of handling it inline cleaner and handles both numerical and complex types uniformly
The JSON format improvement we discussed (flat values list + value_type field) is also in place
The NumericalType import I added to decoder.py is no longer needed the check lives inside ArrayType._is_numerical_array() now


TESTING AND VERIFICATION : 

On the current devel branch:


`pytest test/fprime_gds/common/dp/test_decoder.py -v`

Result :
All 25 tests pass, including the array type tests (test_decode_u8_array, test_decode_u32_array, test_decode_data_array, test_decode_fpp_array, test_decode_array_record).

Note: If you test pre and post merging #292 with the current test files, you won't see much of a performance difference because the test data is very small (~1.9KB). The ~31x speedup shows up on multi-MB data products like the 12MB file from #4784.

Summary
My PR ( this PR #289 ) added the batch deserialization fast path in the decoder. @kubiak-jpl's #292 took that further by moving the optimization into ArrayType itself,  keeping the optimization close to the type system rather than in the decoder. The performance gains are equivalent (~31x speedup on large numerical arrays).

That being said I think we can say good job everyone! and this PR can be closed safely.

thomas-bc:
Ah, I see. Thanks a lot for the detailed assessment, and bummer that you didn't get credit in the form of a commit for this work. Hopefully that's ok.

Good job! And thanks everyone!

PR #290, opened 2026-03-11
Add parse_known_args override to ConfigDrivenParser

ianbrault:
| | |
|:---|:---|

---
## Change Description

This adds a `parse_known_args` override to the `ConfigDrivenParser` class. This functions the same as `ConfigDrivenParser.parse_args` but only parses the known arguments for the provided parser classes.

## Rationale

`ConfigDrivenParser` does not implement `parse_known_args`, it inherits the default implementation from `ParserBase` which does not properly pull in config values. Our use case is a script which ingests the `fprime-gds` config file but only handles a subset of the arguments that it contains.

Use a `ConfigDrivenParser` where the parser classes only handle a subset of arguments that are present in the configuration file.

## Future Work

N/A

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -345,6 +345,42 @@ class ConfigDrivenParser(ParserBase):
             arguments: arguments to process, None to use command line input
         Returns: namespace with all parsed arguments from all provided ParserBase subclasses
         """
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):

thomas-bc:
Thank you!

PR #294, opened 2026-04-11
fix: replace datetime.UTC with timezone.utc for Python 3.10 compatibility

alideniztartma:
Fixes a Python compatibility issue where `datetime.UTC` is used, which is not available in Python 3.10.

Replaced with `datetime.timezone.utc`, which is compatible with Python 3.9+ as required by the project.

Tested locally on Python 3.10.14.

| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#4996 |

---
## Change Description

Replaced datetime.UTC with datetime.timezone.utc in helpers.py. This fixes the specific failure in GDS caused by using datetime.UTC on Python 3.10.

## Rationale

The project's README specifies a Python 3.9+ requirement. However, datetime.UTC was introduced in Python 3.11, causing an AttributeError on Python 3.10. Using datetime.timezone.utc maintains the same functionality while ensuring broad compatibility.

Tested locally on Python 3.10.14. Verified locally that the AttributeError no longer occurs and the sequence uplink process starts successfully in a Python 3.10 environment.
## Future Work

### DIFF
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -30,7 +30,7 @@ jobs:
     steps:
     - name: check-spelling
       id: spelling
-      uses: check-spelling/check-spelling@v0.0.24
+      uses: check-spelling/check-spelling@v0.0.26
       with:
         suppress_push_for_open_pull_request: 1
         checkout: true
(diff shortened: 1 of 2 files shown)


alideniztartma:
It looks like the spell-check workflow is using a version (<0.0.26) that may fail for forked PRs due to restricted permissions. This seems unrelated to the change.

thomas-bc (approved):

PR #295, opened 2026-05-02
Added --skip-browser-open flag

Excaliabure:
| | |
|:---|:---|
|**_Related Issue(s)_**|  #5409 |

---
## Change Description

Added the flag to fprime-gds 
--gui-auto-open False

This disables the main command fprime-gds from opening the gui automatically.

## Rationale

When learning,testing,deploying and what not, it is annoying to have a kabajillion tabs opening just to test 1 function change. Adding this flag allows for having one tab open and simply refreshing upon redeploying, allowing for a smoother work flow imo.

It is also an issue opened on the main fprime issues \#5409

For the most part it should not affect functionality, as it just checks if the flag is true to call webbrowser.open(...) in run_deployment.py. Regardless if this function exists or not the gui still opens on hte port and is acessable 

## Future Work

Perhaps work on the wording of the flag? 
Also change the implimentation from str to bool to make foolproof

## Note 

I chose to use str, as when i tried to use bool, it would not replace the default value with the passed argument. This might be a different issue in itself, however the implentation has no noticable negatives, rather --gui-auto-open f also is a valid response and works in testing, so more of a positive.

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -1287,6 +1287,13 @@ class GdsParser(ParserBase):
                 "type": str,
                 "help": "Set the GUI server address [default: %(default)s]",
             },
+            ("--gui-auto-open",):{
+                "dest": "gui_auto_open",
+                "default": False,
+                "required": False,
+                "type": bool,
+                "help": "Initializes GUI server without auto opening browser [default: %(default)s]"
+                }
         }
 
     def handle_arguments(self, args, **kwargs):
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -1289,9 +1289,10 @@ class GdsParser(ParserBase):
             },
             ("--gui-auto-open",):{
                 "dest": "gui_auto_open",
-                "default": False,
+                "action": "store",
+                "default": "True",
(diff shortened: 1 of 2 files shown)


LeStarch:
I'd recommend we put in a flag of the form:
```
--disable-browser-open`
```

or 
```
--no-browser-open
```

Such that if you want this feature shut-off, you just add the flag.    That way you don't need to remember a "False" argument to a flag.

Definitely a feature we want.

thomas-bc (changes requested):
Agreed with the above comment

See the following example for a similar pattern:

https://github.com/nasa/fprime-gds/blob/62738ce491e43a6f325d61b54eea819f366515ab/src/fprime_gds/common/communication/adapters/uart.py#L175-L180

thomas-bc:
Fixes nasa/fprime#5409

Excaliabure:
Changed it so there is no need for any False/True statement after (empty works). Is --skip-browser-open good?

Current behavior: Currently throws error when calling gds on open port and opens the help menu? 
Suggestion: Ask the user if they want to auto-assign a new open port or wait until the port they want is open.

thomas-bc (approved):

thomas-bc:
LGTM!

PR #301, opened 2026-05-23
fix: Correct sequencer day autocomplete snippet

macayu17:
| | |
  |:---|:---|
  |**_Related Issue(s)_**| nasa/fprime#5194 |

  ---
  ## Change Description

  Corrects the sequencer autocomplete snippet so it inserts `day` instead of `doy`.

  This also adds a focused autocomplete test for the snippet and removes the stale spelling allowlist entry for `doy`.

  ## Rationale

  The GUI suggestion should match the command sequence field name. The current snippet uses `doy`, which is the typo reported in nasa/fprime#5194.

  Ran:

  `python -m pytest test\fprime_gds\flask\static\addons\sequencer\test_autocomplete.py`

  Also checked the patch with:

  `git diff --check HEAD~1..HEAD`

  ## Future Work

  N/A


test/fprime_gds/flask/static/addons/sequencer/test_autocomplete.py
thomas-bc, line 1: Let's remove this file please, I'm not convinced we need a test here.

thomas-bc (changes requested):
Thanks for the quick fix - LGTM but let's remove that test

macayu17:
Done, removed the test file. The PR now only keeps the snippet fix and the spelling allow list cleanup.

src/fprime_gds/flask/static/addons/sequencer/autocomplete.js
```diff
@@ -83,7 +83,7 @@ function autocomplete(commands, node, siblings, state) {
         return autoArg(commands, node, siblings, state);
     }
     else if (siblings.length === 0) {
-        let snips = ["R${hours}:${minutes}:${seconds}", "A${year}-${doy}T${hours}:${minutes}:${seconds}"];
+        let snips = ["R${hours}:${minutes}:${seconds}", "A${year}-${day}T${hours}:${minutes}:${seconds}"];
```
thomas-bc: turns out `doy` was conscious to mean `Day Of Year`. Let's spell it out entirely `day_of_year`

thomas-bc (changes requested):

thomas-bc (approved):

PR #305, opened 2026-05-30
Fixed infinite busy-loop in the fprime GDS TCP server on a socket error + Unbounded client + UDP Handler errors

bitWarrior:
|:---|:---|
|**_Has Unit Tests (y/n)_**| Y |

---
## Change Description

The fprime GDS TCP server has an infinite loop in the recv() function to process packet data. There is a bug processing an OS Exception.  The OSError exception branch only logs the error instead of handling it. So processing falls back into the loop with a broken socket.  This results in a rinse -> cycle -> repeat failure condition since the socket will be broken forever. 

The fix is simple - the code should treat a socket error the same way the code already treats a dead connection

I fixed some other things since I opened the hood on TCP server that are related.
 
1) I switched the packet read accumulator from msg = msg + chunk (rebuilds an immutable bytes every iteration) to a bytearray with .extend(). The is am improvement because "msg = msg + chunk"  builds a brand-new bytes object by copying all the bytes already accumulated, so reassembling an N-byte payload across many chunks copies roughly N² bytes total; a bytearray with .extend() appends in place, making it linear.

2) added MAX_PAYLOAD_SIZE = 64 MiB and a _read_payload_size() helper that validates the 4-byte prefix. A client can no longer declare a ~4 GiB length to force an unbounded read/allocation, and a truncated length prefix no longer throws struct.error. Truncated payloads are now dropped with a warning instead of silently passed downstream. I also added a docstring that informs the user that if their deployment is larger than 64 MiB, they should increase MAX_PAYLOAD_SIZE. This change improves the stability of the f prime GDS, since an unbounded client could cause "undesirable performance" of the GDS.

3) The UDP handler now rejects datagrams shorter than the 9-byte header instead of slicing a runt header.

4) The UDP handler now validates the datagram is long enough for the length prefix instead of truncating it and not telling the user. I also added guards in the header split so a malformed header logs and returns instead of crashing. Not only does this help debugging new deployments, but prevents the GDS from processing malformed data.

## Rationale

Infinite loop failure condition and improve the robustness of the GDS during off-nominal conditions.

I created a new unit test for tcpserver that tests and verifies the fixes and robustness I described above.

## AI Usage

I had claude implement the code changes and create the new tcpserver unit test.

## Future Work

None.

### DIFF
--- a/src/fprime_gds/executables/tcpserver.py
+++ b/src/fprime_gds/executables/tcpserver.py
@@ -197,19 +197,23 @@ class ThreadedTCPRequestHandler(socketserver.StreamRequestHandler):
 
     def recv(self, l):
         """
-        Read l bytes from socket.
+        Read exactly l bytes from the socket.
+
+        Returns the l bytes on success, or b"" if the connection is closed or
+        broken before l bytes arrive. Callers treat b"" as "client gone".
         """
-        chunk = b""
-        msg = b""
-        n = 0
-        while l > n:
+        # Accumulate into a bytearray instead of rebuilding an immutable bytes
+        # object each iteration. "msg = msg + chunk" is O(n^2) and, combined
+        # with an unbounded length field, lets a client drive quadratic CPU and
+        # memory use on a large read.
+        msg = bytearray()
+        while len(msg) < l:
             try:
-                chunk = self.request.recv(l - n)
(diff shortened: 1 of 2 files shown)


thomas-bc (changes requested):
CI failures

LeStarch:
Closing as stale.

PR #312, opened 2026-06-10
Fix ConstantJsonLoader crash when constants field is missing

ryanhasmanners:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/5252 |

---
## Change Description

Updates `ConstantJsonLoader` to handle a missing `constants` field by defaulting to an empty list rather than attempting to access a non-existent dictionary key.

## Rationale

`ConstantJsonLoader` currently emits a warning when the constants field is missing, but then still attempts to access the missing field, raising a `KeyError`.

This change aligns the implementation with the existing warning behaviour by treating missing constants as an empty collection and allowing startup to continue.

1. Load a JSON dictionary containing a valid constants field and verify existing behaviour is unchanged.
2. Load a JSON dictionary without a constants field and verify:
    - A warning is emitted.
    - No exception is raised, and empty constants dictionary is returned.

## Future Work

None

### DIFF
--- a/src/fprime_gds/common/loaders/constant_json_loader.py
+++ b/src/fprime_gds/common/loaders/constant_json_loader.py
@@ -31,10 +31,11 @@ class ConstantJsonLoader(JsonLoader):
         """
         name_dict = {}
 
+        constants = self.json_dict.get(self.CONSTANTS_FIELD, [])
         if self.CONSTANTS_FIELD not in self.json_dict:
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):

thomas-bc:
Thank you very much for coming in to fix this! 🚀

PR #313, opened 2026-06-10
Port sequence parser to lark

Kronos3:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Ports the manual sequence parser to a lark grammar to make it easier to extend with more complex expressions. I also added additional unit tests to get the coverage up on this and the old parser. The tests pass on both (ignoring some delta in parsing error strings).

## Rationale

We want to add array/struct expressions to seqgen which requires a more complete parsing solution..

N/A

## Future Work

Add struct/array expressions to the grammar.


thomas-bc (approved):
That's super cool! 🚀 

Also as a note if you prefer to switch to pytest in your future tests that would also work, but since the others in that file use unittest this is good

PR #315, opened 2026-06-15
Bound the comm downlink queue

peter941221:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#5225 |

---
## Change Description

This fixes the unbounded downlink queue in `Downlinker`.

Before this patch, `Downlinker` built `self.outgoing = Queue()` with no `maxsize`, so `put_nowait()` never raised `Full`. That left both warning paths unreachable:

- `GDS ground queue full, dropping frame`
- `GDS ground queue full, dropping loopback frame`

This patch now does four things:

1. sets the default ground queue to `4096` frames
2. exposes that bound through `--downlink-queue-maxsize` and the existing config file `command-line-options` path
3. moves `Full` handling inside the deframing loop so each frame is handled independently once the queue fills
4. throttles queue-full warnings to one message per saturation episode, then re-enables the warning once the queue has room again

The per-frame handling matters because once one `put_nowait()` hits `Full`, the remaining frames in the same deframed batch should still follow the same drop path one by one instead of aborting the whole batch loop at the first exception.

The warning throttle matters because otherwise a sustained full queue trades a RAM growth problem for an unbounded log growth problem.

## Rationale

The existing code already chose a drop-on-full policy. The bug was that the queue implementation never let that policy execute.

Bounding the queue restores the intended backpressure boundary. Handling `Full` per frame keeps saturation behavior local to each frame. Throttling the warning keeps the logging path bounded under sustained pressure.

Ran:

`PYTHONPATH=src python -m pytest test/fprime_gds/common/communication/test_updown.py test/fprime_gds/executables/test_comm_cli.py -q`

Result:

`8 passed`

Coverage here checks:

1. the bounded default queue size
2. CLI parsing for `--downlink-queue-maxsize`
3. config-file parsing for `downlink-queue-maxsize`
4. loopback frame drops when the queue is full
5. deframing keeps the first available slot, then drops later frames in the same batch once the queue fills
6. repeated queue-full drops only emit one warning while saturation persists
7. the warning is emitted again after the queue has room and then fills again


LeStarch:
For this to be accepted, we need two things:
1. The queue depth must be configurable
2. We must bound the number of logged warning message.  Otherwise we change a RAM overrun to a FS overrun (which is much more dangerous).

peter941221:
Configurable queue depth is in now: `--downlink-queue-maxsize` wires through the existing config/CLI path, and the downlink drop warning is throttled to one message per saturation episode. I added focused tests for the default bound, CLI/config parsing, and warning reset after recovery.

PR #325, opened 2026-07-08
Add configurable GDS application root for reverse-proxy deployments

Abhishek21g:
## Summary

Implements configurable path-prefix support for the GDS web UI and API, addressing nasa/fprime#3854 (CCB-approved by @LeStarch).

- Add `FP_GDS_APPLICATION_ROOT` env var and `--gui-root-path` CLI flag
- Serve static assets and API routes under the configured prefix via `ScriptNameMiddleware`
- Inject `bootstrap-config.js` so the browser prefixes fetch URLs correctly
- Fix `ui_url` printed at startup to include the prefix

## Test plan

- [x] `pytest test/fprime_gds/flask/test_prefix.py` (4 tests)
- [ ] Manual: `fprime-gds --gui-root-path /fprime-gds-2` behind a reverse proxy
- [ ] Manual: verify default behavior unchanged when prefix is unset

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -1296,7 +1296,15 @@ class GdsParser(ParserBase):
                 "dest": "browser_auto_open",
                 "action": "store_false",
                 "help": "Run server without auto-launching the default web browser"
-                }
+                },
+            ("--gui-root-path",): {
+                "dest": "gui_root_path",
+                "action": "store",
+                "default": "",
+                "required": False,
+                "type": str,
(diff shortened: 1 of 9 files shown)


LeStarch:
This change adds several things that are beyond the scope of that ticket:
1. It passes the prefix to JavaScript
2. It adds in handling for proxies that don't strip the changed path.

All this ticket approved was to allow the GDS to run at a path specified url path offset.

We should:
1. Implement a native JavaScript solution for detecting the offset and not require passing the prefix.
2. Remove the .wsgi_app override.  For now, we should not need to support this case.

thomas-bc:
Closing for staleness and per ticket description this is not in plan for now

PR #328, opened 2026-07-16
Preserve literal percent placed directly before a conversion in C-style format strings

arpitjain099:
| | |
|:---|:---|

---
## Change Description

`preprocess_c_style_format_str` in `common/utils/string_util.py` converts C-style format strings to Python format strings. The pattern begins with `(?<!%)(?:%%)*%...`, so any run of literal `%%` pairs sitting right before a conversion is swallowed by that non-capturing group. `convert()` only rebuilt the field from the flags/width/precision/type groups, so those leading pairs were never re-emitted and the literal percent was lost.

Concretely, `%%%d` was turned into `{}` instead of `%{}`, and the full pipeline `format_string_template(preprocess_c_style_format_str("%%%d"), 5)` returned `"5"` where C prints `"%5"`.

The fix captures the `%%` run and puts it back in front of the converted field. The existing trailing `.replace("%%", "%")` still collapses the pairs to single percents uniformly, so no other case changes.

## Rationale

Fixes a correctness bug: a literal percent that immediately precedes a formatted value silently disappears from event and channel display text. Format strings like `"progress %%%d complete"` rendered as `"progress 50 complete"` instead of `"progress %50 complete"`.

- `pytest test/fprime_gds/common/utils/test_string_util.py` (28 passed).
- Two tests were added, `test_format_percent_sign_before_conversion` and `test_format_percent_sign_before_conversion_in_text`. Both fail on `devel` and pass with this change.
- Worth confirming the unchanged cases still hold: standalone `%%` (`"%.2f%%"` -> `"1.23%"`), a plain `%d`, and text like `"100%% done: %d"`.

## Future Work

None.

### DIFF
--- a/src/fprime_gds/common/utils/string_util.py
+++ b/src/fprime_gds/common/utils/string_util.py
@@ -89,7 +89,9 @@ def preprocess_c_style_format_str(format_str: str) -> str:
     def convert(match_obj: re.Match):
         if match_obj.group() is None:
             return match_obj
-        flags, width, precision, _, conversion_type = match_obj.groups()
+        literal_percents, flags, width, precision, _, conversion_type = (
+            match_obj.groups()
+        )
         format_template = ""
         if flags:
             format_template += f"{flags}"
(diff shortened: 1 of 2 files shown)


thomas-bc:
This was not requested because C-style format strings are not used anymore. This is a remnant of the XML days. We don't do development there anymore. Closing.

PR #330, opened 2026-07-27
Fix fprime-prm-write dropping parameters with falsy values

tekinertekin:
| | |
|:---|:---|
|**_Related Issue(s)_**|  nasa/fprime#5451 |

---
## Change Description

parse_json() used `if not prm_val` to skip parameters that have no value, which also skipped parameters explicitly set to a falsy value (0, 0.0, "", False) -- silently omitting them from the output. An explicit falsy value overriding a truthy default dropped the parameter entirely. Test with `is None` so only genuinely absent values are skipped and explicit falsy values are preserved. AI-assisted development; I understand and take responsibility for every line.

## Rationale

Refs nasa/fprime#5451

Added a regression test covering falsy parameter values.

## Future Work

None.

### DIFF
--- a/src/fprime_gds/common/tools/params.py
+++ b/src/fprime_gds/common/tools/params.py
@@ -158,7 +158,7 @@ def parse_json(param_value_json, name_dict: dict[str, PrmTemplate], include_impl
                 # get the value
                 prm_val = comp_json[prm_template.prm_name]
         
-        if not prm_val:
+        if prm_val is None:
             # not writing a val for this prm
             continue
 
--- a/test/fprime_gds/common/tools/test_paramdb_gen.py
+++ b/test/fprime_gds/common/tools/test_paramdb_gen.py
@@ -1,7 +1,10 @@
 import filecmp
 from pathlib import Path
+import pytest
 import tempfile
-from fprime_gds.common.tools.params import convert_json
+
+from fprime_gds.common.tools.params import convert_json, parse_json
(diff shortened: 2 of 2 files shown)


beavuhs:
Can confirm this is a bug - I can locally reproduce this with boolean and integer parameters that I try to write via a .json.

tekinertekin:
> Can confirm this is a bug - I can locally reproduce this with boolean and integer parameters that I try to write via a .json.

Thanks for confirming. Booleans are a good catch — `False` was not in the
parametrized test, only `""`, `0` and `0.0`. I have added it; it fails the
same way without the fix (`assert 'FAKE_PRM' in []`) and passes with it.

durasi:
I opened #331 for the same parse_json bug before I spotted this one; yours is the earlier and more focused PR, so I'll close mine in favor of it.

One case from my tests that might be worth folding in here: with include_implicit=True, an explicit falsy value should override the dictionary default rather than fall back to it. A quick check that pins the returned value (not just the parameter's presence) would cover it:

`def test_explicit_falsy_overrides_default(name_dict):
    fqn, template = next(iter(name_dict.items()))
    result = parse_json({template.comp_name: {template.prm_name: 0}}, name_dict, True)
    matches = [val for tmpl, val in result if tmpl is template]
    assert matches == [0]`
    
Feel free to use or skip it. Either way, thanks for the fix.

LeStarch (approved):

thomas-bc (approved):

PR #331, opened 2026-08-02
Fix fprime-prm-write dropping falsy parameter values

durasi:
## Change Description
`parse_json` decided whether a parameter had a value with `if not prm_val`, which silently drops explicit falsy values (`0`, `0.0`, `false`, `""`). Changed to an explicit `is None` check and added regression tests covering each falsy value plus the implicit-defaults override case.

## Rationale
Addresses nasa/fprime#5451. A parameter file that silently omits explicitly-set values is dangerous for operations: the tool reports success while the uplinked file leaves the parameter at its previous value.

## Testing
New tests in `test/fprime_gds/common/tools/test_prm_falsy.py`; all pass locally alongside the existing `test_prm_decode.py` suite.

### DIFF
--- a/src/fprime_gds/common/tools/params.py
+++ b/src/fprime_gds/common/tools/params.py
@@ -158,8 +158,10 @@ def parse_json(param_value_json, name_dict: dict[str, PrmTemplate], include_impl
                 # get the value
                 prm_val = comp_json[prm_template.prm_name]
         
-        if not prm_val:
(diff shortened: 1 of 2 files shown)


durasi:
Closing in favor of #330, which fixes the same parse_json bug, was opened earlier, and has already been independently reproduced. I've left my default-override test case as a suggestion there. Thanks @tekinertekin.

PR #332, opened 2026-08-04
Add --gds-non-clearing-history to keep history for late-connecting clients

### LINKED ISSUE #3604: Support not clearing history in F Prime GDS
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| GDS |
---
## Feature Description

When the web browser interface is loaded, it will not display any previously received data. This is a problem because it allows us to miss the first few events emitted by a target application, depending on a race between establishing communication with the application and opening the webapp in the browser.

In my local copy of F Prime GDS, I was able to work around this by disabling the automatic clearing feature and changing the default cursor position. I could do that by using this pipeline history implementation in `fprime_gds.flask.components`:

```
class NonClearingHistory(FlaskEndpointRamHistory):
    def retrieve(self, start=None, limit=None):
        # Start from the beginning
        with self.lock:
            if start not in self.retrieved_cursors:
                self.count_offsets[start] = self.count
                self.retrieved_cursors[start] = 0
            return super().retrieve(start, limit)

    def clear(self, start=None):
        # Do not clear history
        pass
```

However, since there is no way to inject a plugin into the flask process, I have to carry this patch.

I would like to request that we either:

1. Support this as a first class feature, gated by one or more command line options.
2. Support a plugin architecture that would let me inject this feature when my team runs the F Prime GDS.

## Rationale

It is surprising and confusing for important events to not be visible, just because the web app didn't open immediately.

Moferanoluwa:
| | |
|:---|:---|
|**_Related Issue(s)_**| Fixes nasa/fprime#3604 |

---
## Change Description

Adds an opt-in `--gds-non-clearing-history` flag that keeps the full history buffer around for the lifetime of the process, instead of the default behavior where a client connecting after data has already arrived misses it.

## Rationale

I hit the same thing the issue describes: the web UI missing the first few events because of the race between the deployment starting and the browser tab opening. The root cause is in `RamHistory.retrieve()` — a brand new session's cursor defaults to the current end of history, so anything that arrived before that session existed is gone as far as that client is concerned. The issue reporter had already worked out a local patch for this (a `NonClearingHistory` override), so I built on that rather than starting from scratch, and wired it up as a first-class CLI flag per the direction on the issue.

`HistoryParser` is a new small parser added to `StandardPipelineParser`'s constituents, which means the flag is picked up by the top-level `fprime-gds` command and carried through to the Flask subprocess for free — `run_deployment.py`'s `launch_html()` already reproduces `StandardPipelineParser`'s arguments when it spawns the Flask process, so I didn't need to add any extra plumbing there.

`NonClearingHistory` starts new sessions at index 0 instead of the current end, and makes `clear()` a no-op. I want to flag one thing I had to work through carefully: a naive version that only overrode the cursor broke, because `FlaskEndpointRamHistory.retrieve()` re-checks `retrieved_cursors` membership for its own count-tracking bookkeeping, and it would silently skip that if I'd already set the cursor first. Setting `count_offsets` alongside the cursor fixed it — worth a close look in review since it's the subtlest part of this change.

Added `test/fprime_gds/flask/test_components.py`:
- `HistoryParser` defaults to off and round-trips the flag correctly through `ParserBase.parse_args`.
- `NonClearingHistory`: a new session sees full pre-existing history, `clear()` doesn't evict anything, and a session that's already caught up only sees genuinely new items (not a replay).
- A regression check that plain `FlaskEndpointRamHistory` is unaffected — new sessions still only see what arrives after they connect.

Ran the full suite before and after (via `git stash`) to make sure I wasn't missing something: 322 passed with this change (316 without, plus my 6 new tests), and the same 30 failures / 6 errors exist either way — pre-existing, not something this touches.

## Future Work

None anticipated.

### DIFF
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -1188,6 +1188,30 @@ class FileHandlingParser(ParserBase):
         return args
 
 
+class HistoryParser(ParserBase):
+    """Parser for the pipeline's in-memory history behavior"""
+
+    DESCRIPTION = "History options"
+
+    def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
+        """Arguments controlling how the pipeline's history is retained"""
+        return {
+            ("--gds-non-clearing-history",): {
+                "dest": "non_clearing_history",
+                "action": "store_true",
+                "default": False,
+                "help": "Do not clear history as it is retrieved by GDS clients (e.g. the web UI). By "
+                "default, history is cleared once seen so a client connecting after data has "
+                "already arrived (a race between the deployment starting and the UI loading) will "
+                "miss it. Setting this retains all history for the lifetime of the process.",
+            },
+        }
+
+    def handle_arguments(self, args, **kwargs):
+        """Handle arguments as parsed"""
+        return args
+
+
 class StandardPipelineParser(CompositeParser):
     """Standard pipeline argument parser: combination of MiddleWare and"""
 
@@ -1197,6 +1221,7 @@ class StandardPipelineParser(CompositeParser):
         FileHandlingParser,
         MiddleWareParser,
         LogDeployParser,
+        HistoryParser,
     ]
 
     def __init__(self):
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -1196,8 +1196,8 @@ class HistoryParser(ParserBase):
     def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
         """Arguments controlling how the pipeline's history is retained"""
         return {
-            ("--gds-non-clearing-history",): {
-                "dest": "non_clearing_history",
+            ("--no-clear-history",): {
+                "dest": "no_clear_history",
                 "action": "store_true",
                 "default": False,
                 "help": "Do not clear history as it is retrieved by GDS clients (e.g. the web UI). By "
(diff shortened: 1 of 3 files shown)


src/fprime_gds/executables/cli.py
```diff
@@ -1188,6 +1188,30 @@ def handle_arguments(self, args, **kwargs):
...
 
+class HistoryParser(ParserBase):
+    """Parser for the pipeline's in-memory history behavior"""
+
+    DESCRIPTION = "History options"
+
+    def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
+        """Arguments controlling how the pipeline's history is retained"""
+        return {
+            ("--gds-non-clearing-history",): {
```
celskeggs: I suggest naming the argument `--no-clear-history` to be concise.

test/fprime_gds/flask/test_components.py
celskeggs, line 1: These tests do not validate that the command line argument is interpreted correctly. They would still pass even if the change to the `pipeline.histories.implementation` initialization was nonfunctional.

celskeggs (changes requested):
Thanks for working on this! I'm not on the F Prime core team, so I can't actually approve your PR, but I think this looks good; I just have a couple of notes after looking through it.

Moferanoluwa:
Both good calls, thanks. Renamed to --no-clear-history, and split the class-selection logic out of setup_pipelined_components into its own select_history_implementation() function so I could test the actual CLI-arg-to-class path directly - the previous tests would have missed a break there since they tested the parser and the history class separately, never the connection between them.

celskeggs (approved):
Looks great to me!

LeStarch (approved):

PR #340, opened 2026-08-12
Add configurable base URL support to the GDS

sylvesterkaczmarek:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#3854 |

---
## Change Description

Adds a `--base-url` option for serving the HTML GDS and REST API below a path prefix such as `/mission/gds`.

- normalizes and validates path-only base URLs
- passes the configured prefix to the Flask subprocess and browser-launch URL
- prefixes Flask and flask-restful routes while preserving existing root behavior by default
- makes browser API requests relative to the loaded UI path so reverse-proxy prefixes are retained
- keeps the image-display download URL base-relative
- adds unit coverage and user documentation

## Rationale

This implements the CCB-approved request in nasa/fprime#3854 and allows multiple GDS instances or reverse-proxy deployments to share a host without requiring a separate port for every GDS.

Run `pytest test/fprime_gds/executables/test_base_url.py` and the normal repository checks. Review URL normalization, Flask route prefixing, and preservation of root-path behavior when `--base-url` is omitted.

## Future Work

None planned for this issue.

### DIFF
--- a/docs/base-url.rst
+++ b/docs/base-url.rst
@@ -1,0 +1,18 @@
+Serving the GDS below a base URL
+================================
+
+The HTML GDS can be served below a URL path instead of the web-server root. This is useful when the GDS is placed behind a reverse proxy or when multiple GDS instances share one host.
+
+Use ``--base-url`` when starting the GDS::
+
+   fprime-gds --base-url /mission/gds
+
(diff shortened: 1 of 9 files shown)


thomas-bc:
Closing - per ticket description this is not something we have the bandwidth to get to

PR #349, opened 2026-08-31
Fix event decoder corrupting events after the first in a multi-event buffer

HuzaifaAbdulRehman:
| | |
|:---|:---|
|**_Related Issue(s)_**| #4620 |

---
## Change Description

`EventDecoder.decode_api` decodes a buffer that may hold several events, looping over each one the way `ChannelDecoder` loops over a multi-item telemetry packet. It advanced past each event's arguments with:

```python
(size, arg_vals) = self.decode_args(data, ptr, event_temp)
ptr += size
```

`decode_args` returns the absolute offset one past the arguments, not a length, so `ptr += size` jumped roughly a whole event too far. Every event after the first was then read from the wrong offset and came back corrupted, or raised `DecodingException` when the stray bytes did not match a known event id. The fix sets `ptr` to the returned offset. It also rewrites the `decode_args` docstring, which never stated that its first return value is that offset, to document the contract whose misreading caused the bug.

## Rationale

Addresses #4620, where events arriving close together came back duplicated or garbled through `send_and_await_event`. A buffer with a single event decodes correctly, which is why the bug stayed hidden.

Added `test/fprime_gds/common/decoders/test_event_decoder.py`: a single-event case and a three-event buffer. On `devel` the multi-event case fails (`Event 23351023 not found in dictionary`, the second event's argument read as an id) and passes with this change. Run `pytest test/fprime_gds/common/decoders`. Other suite failures (history, seq_parser, plugins) are present on unmodified `devel`. Verified on Python 3.10; CI covers 3.9 to 3.13.

## Future Work

None.

### DIFF
--- a/src/fprime_gds/common/decoders/event_decoder.py
+++ b/src/fprime_gds/common/decoders/event_decoder.py
@@ -88,11 +88,11 @@ class EventDecoder(decoder.Decoder):
 
             event_temp = self.__dict[event_id]
 
-            (size, arg_vals) = self.decode_args(data, ptr, event_temp)
+            (arg_end_offset, arg_vals) = self.decode_args(data, ptr, event_temp)
 
             event_list.append(event_data.EventData(arg_vals, event_time, event_temp))
-            # add up argument sizes
-            ptr += size
+            # decode_args returns the absolute offset past the arguments; advance to it
+            ptr = arg_end_offset
         return event_list
 
     @staticmethod
@@ -110,10 +110,12 @@ class EventDecoder(decoder.Decoder):
                       arg_data goes to.
 
         Returns:
-            Parsed arguments in a tuple (order the same as they were parsed in).
-            Each element in the tuple is an instance of the same class as the
-            corresponding arg_type object in the template parameter. Returns
-            none if the arguments can't be parsed
+            A two element list ``[end_offset, args]``. ``end_offset`` is the
+            absolute offset into ``arg_data`` one past the last argument; the
+            caller uses it as the start of the next event in the buffer. ``args``
+            is a tuple of the parsed arguments in order, each an instance of the
+            same class as the corresponding arg_type object in the template.
+            Raises DecodingException if an argument cannot be parsed.
         """
         arg_results = []
         args = template.get_args()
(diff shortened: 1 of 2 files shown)


thomas-bc (changes requested):
While this does fix something that looks buggy, I don't see how it fixes the mentioned issue?

HuzaifaAbdulRehman:
You're right, this doesn't fix #4620. In the normal pipeline the distributor hands the event decoder one event per `decode_api` call, so the multi-event loop this patch corrects is never reached. #4620 is in the sequence search instead: `find_history_sequence` reads with `retrieve_new(repeats=True)`, and when events arrive out of FSW-time order, `data_callback` moves the cursor back (`min(index, retrieved_cursor)`). An already-matched event is then re-delivered, and with identical predicates (`["MyInfoEvent"] * 4`) the next one matches it again, duplicating that event and dropping the real last one. I can open a separate fix for that, though it touches the repeats semantics so I would confirm the intended behavior with you first. Do you want me to close this one, or keep it as a standalone decoder fix?

thomas-bc:
I'm not sure I follow your PR description then

> Addresses #4620

Why ?

HuzaifaAbdulRehman:
It shouldn't say that, the line is wrong. I wrote the description before I traced the pipeline and didn't correct it once I realized the decoder path isn't reached. Closing this rather than leave a fix mislabeled to an issue it doesn't touch. If a sequence-search fix for #4620 would be useful, I'll open that separately.

HuzaifaAbdulRehman:
Follow-up now that I've traced the real cause of #4620, which is in the sequence search rather than the decoder. `find_history_sequence` reads with `retrieve_new(repeats=True)`, so when events arrive out of FSW-time order `ChronologicalHistory` re-sorts them and re-delivers ones that already matched an earlier position. With identical predicates (`[evt] * 4`) the searcher then matches the same event twice, duplicating it and dropping the real last one. I reproduced that exact symptom through `find_history_sequence` on `devel` and have a small fix (skip an already-matched item) with a regression test that goes red without it. Before I open a PR: would you want the fix in the searcher, or in the cursor/repeats logic? Happy to go whichever way you prefer.

PR #354, opened 2026-09-12
Updated fprime_test_api to pickup config in FPRIME_GDS_CONFIG_PATH

bevinduckett:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Has Unit Tests (y/n)_**|  n |

---
## Change Description

Updated files for the fprime_test_api pytext fixture to use the configuration from the file in the FPRIME_GDS_CONFIG_PATH environment variabl

## Rationale

fprime-gds will pick up default configuration from the file in the FPRIME_GDS_CONFIG_PATH environment variable (if that's set). It would make using fprime_test_api simpler if configuration for that tool also picked up the options from the file in FPRIME_GDS_CONFIG_PATH

Specify configuration (e.g. see example below)  in a file , then set environment variable FPRIME_GDS_CONFIG to point to that. Then start fprime-gds and verify it picks up the correct option and then run a test that uses fprime_test_api and verify it picks up those options 

Below is an example of the config file we're using (with dictionary path anonymized)
```
command-line-options:
  communication-selection: jlink-comm
  dictionary: /workspaces/project/build-artifacts/va416x0-baremetal/DemoDeployment/dict/DemoDeployment.json
  framing-selection: fprime
  jlink-speed: 2000
  no-app:
  no-clear-history:
  no-zmq:
```

## Future Work

### DIFF
--- a/src/fprime_gds/common/testing_fw/pytest_integration.py
+++ b/src/fprime_gds/common/testing_fw/pytest_integration.py
@@ -1,4 +1,4 @@
-""" pytest_integration.py: F´ fixture API fixture
+"""pytest_integration.py: F´ fixture API fixture
 
 pytest uses fixtures to provide for extra functionality when writing tests. This fixture sets up the F´ stack allowing
 our tests to use a configured test API without needing to create any specific setup to use that API. i.e. write a test
@@ -14,19 +14,20 @@ Here a test (defined by starting the name with test_) uses the fprime_test_api f
 
 @author lestarch
 """
+
 import itertools
 import sys
 from pathlib import Path
 import pytest
 
 from fprime_gds.common.testing_fw.api import IntegrationTestAPI
-from fprime_gds.executables.cli import StandardPipelineParser
+from fprime_gds.executables.cli import StandardPipelineParser, ConfigDrivenParser
 
 SEQUENCE_COUNTER = itertools.count()
 
 
 def pytest_addoption(parser):
-    """ Add fprime-gds options to the pytest parser
+    """Add fprime-gds options to the pytest parser
 
     Pytest allows users to add options to its parser. These options act very similar to argparse options and thus can be
     reused from the standard GDS command line processing. Note: pytest restricts the use of short flags (-[a-z]) thus we
@@ -39,7 +40,7 @@ def pytest_addoption(parser):
         # Reduce flags to only the long option (i.e. --something) form
         flags = [flag for flag in flags if flag.startswith("--")]
         parser.addoption(*flags, **specifiers)
-        
+
     # Add an option to specify JUnit XML report file
     parser.addoption(
         "--junit-xml-file",
@@ -51,7 +52,7 @@ def pytest_addoption(parser):
     parser.addoption(
         "--gen-junitxml",
         action="store_true",
-        help="Enable JUnitXML report generation to a specified location"
+        help="Enable JUnitXML report generation to a specified location",
     )
     # Add an option to specify a json config file that maps components
     parser.addoption(
@@ -64,19 +65,20 @@ def pytest_addoption(parser):
     parser.addoption(
         "--use-yamcs",
         action="store_true",
-        help="Use YAMCS transport instead of TCP socket"
+        help="Use YAMCS transport instead of TCP socket",
     )
     parser.addoption(
         "--yamcs-url",
         action="store",
         default="http://localhost:8090",
-        help="YAMCS server URL [default: %(default)s]"
+        help="YAMCS server URL [default: %(default)s]",
     )
 
+
 def pytest_configure(config):
-    """ This is a hook to allow plugins and conftest files to perform initial configuration
-    
-    This hook is called for every initial conftest file after command line options have been parsed. After that, the 
+    """This is a hook to allow plugins and conftest files to perform initial configuration
+
+    This hook is called for every initial conftest file after command line options have been parsed. After that, the
     hook is called for other conftest files as they are registered.
     """
     # Create a JUnit XML report file to capture the test result in a specified location
@@ -86,9 +88,10 @@ def pytest_configure(config):
             raise pytest.UsageError("--gen-junitxml requires --logs to be specified")
         config.option.xmlpath = Path(logs) / config.getoption("--junit-xml-file")
 
-@pytest.fixture(scope='session')
+
+@pytest.fixture(scope="session")
 def fprime_test_api_session(request):
-    """ Create a session-level fprime test API
+    """Create a session-level fprime test API
 
     This is a pytest session fixture. Using the options added above, this will parse the necessary options for
     connecting the standard pipeline to the running GDS. This pipeline is supplied to the fprime test API returned as
@@ -106,11 +109,24 @@ def fprime_test_api_session(request):
         fprime test API connected to the GDS.  Note: a second call will shut down that object.
     """
     pipeline_parser = StandardPipelineParser()
+
+    # Use the ConfigDrivenParser to retrieve default configuration from a file (so that pytest
+    # behavior matches fprime-gds CLI behavior). ConfigDrivenParser.parse_known_args() can NOT
+    # be called with arguments=None here, as that defaults to sys.argv[1:] which is pytest's own
+    # command line (e.g. -v, --color=yes) and not fprime-gds options. Instead, reproduce only the
+    # standard-pipeline flags that pytest actually parsed explicitly, and let ConfigDrivenParser
+    # fill in the rest from the configuration file.
+    reproduced_args = pipeline_parser.reproduce_cli_args(
+        request.config.known_args_namespace
+    )
+    arg_ns, _, _ = ConfigDrivenParser.parse_known_args(
+        [StandardPipelineParser], arguments=reproduced_args, client=True
+    )
+
     pipeline = None
     api = None
     deployment_config = None
     try:
-        arg_ns = pipeline_parser.handle_arguments(request.config.known_args_namespace, client=True)
 
         if request.config.getoption("--use-yamcs"):
             try:
@@ -154,9 +170,9 @@ def fprime_test_api_session(request):
             print(f"[WARNING] Exception in pipeline teardown: {exc}", file=sys.stderr)
 
 
-@pytest.fixture(scope='function')
+@pytest.fixture(scope="function")
 def fprime_test_api(fprime_test_api_session, request):
-    """ Provide a per-testcase fixture
+    """Provide a per-testcase fixture
 
     Although the test API should exist across all testcases and thus be created at the "session" level, individual test
     cases need a clean test API that has logged the testcases has started. Thus, the session API is refined into a
--- a/src/fprime_gds/executables/cli.py
+++ b/src/fprime_gds/executables/cli.py
@@ -38,7 +38,6 @@ from fprime_gds.plugin.definitions import PluginType
 from fprime_gds.plugin.system import Plugins, PluginsNotLoadedException
 from fprime_gds.common.zmq_transport import ZmqClient
 
-
 GUIS = ["none", "html"]
 
 
@@ -55,7 +54,11 @@ class ParserBase(ABC):
     @property
     def description(self) -> str:
         """Return parser description"""
-        return self.DESCRIPTION if self.DESCRIPTION is not None else "Unknown command line parser"
+        return (
+            self.DESCRIPTION
+            if self.DESCRIPTION is not None
+            else "Unknown command line parser"
+        )
 
     @abstractmethod
     def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
@@ -318,14 +321,30 @@ class ConfigDrivenParser(ParserBase):
 
     DEFAULT_CONFIGURATION_PATH = Path("fprime-gds.yml")
 
+    # Takes precedence over DEFAULT_CONFIGURATION_PATH
+    DEFAULT_CONFIGURATION_PATH_ENV = "FPRIME_GDS_CONFIG_PATH"
+
     @classmethod
     def set_default_configuration(cls, path: Path):
         """Set path for (global) default configuration file
 
         Set the path for default configuration file. If unset, will use 'fprime-gds.yml'. Set to None to disable default
-        configuration.
+        configuration. Calling this function disables the environment variable override.
         """
         cls.DEFAULT_CONFIGURATION_PATH = path
+        os.environ.pop(cls.DEFAULT_CONFIGURATION_PATH_ENV, None)
+
+    @classmethod
+    def get_default_configuration(cls):
+        """Get path for (global) default configuration file
+
+        If set, the environment variable (DEFAULT_CONFIGURATION_PATH_ENV) overrides
+        DEFAULT_CONFIGURATION_PATH. Get the path for default configuration file. If unset, will
+        use 'fprime-gds.yml'.
+        """
+        if cls.DEFAULT_CONFIGURATION_PATH_ENV in os.environ:
+            return Path(os.environ[cls.DEFAULT_CONFIGURATION_PATH_ENV])
+        return cls.DEFAULT_CONFIGURATION_PATH
 
     @classmethod
     def parse_args(
@@ -450,7 +469,7 @@ class ConfigDrivenParser(ParserBase):
             ("-c", "--config"): {
                 "dest": "config",
                 "required": False,
-                "default": self.DEFAULT_CONFIGURATION_PATH,
+                "default": self.get_default_configuration(),
                 "type": Path,
                 "help": "Argument configuration file path. [default: %(default)s]",
             },
@@ -969,7 +988,9 @@ class LogDeployParser(ParserBase):
                 args.log_prefix = tool_name
 
             timestamp = datetime.datetime.now().strftime("%Y_%m_%d-%H_%M_%S")
-            dir_name = f"{args.log_prefix}-{timestamp}" if args.log_prefix else timestamp
+            dir_name = (
+                f"{args.log_prefix}-{timestamp}" if args.log_prefix else timestamp
+            )
             args.logs = os.path.abspath(os.path.join(args.logs, dir_name))
             # A dated directory has been set, all log handling must now be direct
             args.log_directly = True
@@ -1175,7 +1196,6 @@ class FileHandlingParser(ParserBase):
                 "type": str,
                 "help": "Directory to store uplink and downlink files. Default: %(default)s",
             },
-
             ("--file-uplink-cooldown",): {
                 "dest": "file_uplink_cooldown",
                 "action": "store",
@@ -1334,11 +1354,11 @@ class GdsParser(ParserBase):
                 "type": str,
                 "help": "Set the GUI server address [default: %(default)s]",
             },
-            ("--skip-browser-open",):{
+            ("--skip-browser-open",): {
                 "dest": "browser_auto_open",
                 "action": "store_false",
-                "help": "Run server without auto-launching the default web browser"
-                }
+                "help": "Run server without auto-launching the default web browser",
+            },
         }
 
     def handle_arguments(self, args, **kwargs):


src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -106,11 +109,24 @@ def fprime_test_api_session(request):
...
+    # Use the ConfigDrivenParser to retrieve default configuration from a file (so that pytest
+    # behavior matches fprime-gds CLI behavior). ConfigDrivenParser.parse_known_args() can NOT
+    # be called with arguments=None here, as that defaults to sys.argv[1:] which is pytest's own
+    # command line (e.g. -v, --color=yes) and not fprime-gds options. Instead, reproduce only the
+    # standard-pipeline flags that pytest actually parsed explicitly, and let ConfigDrivenParser
+    # fill in the rest from the configuration file.
+    reproduced_args = pipeline_parser.reproduce_cli_args(
+        request.config.known_args_namespace
+    )
+    arg_ns, _, _ = ConfigDrivenParser.parse_known_args(
```
thomas-bc-autowot, line 122: [Test Quality] **must fix** New config-file loading in the `fprime_test_api_session` fixture has no test, and a test exposes that it does not work for defaulted options.

`request.config.known_args_namespace` already carries argparse defaults, so `reproduce_cli_args` emits e.g. `--tts-port 50050 --tts-addr 0.0.0.0 --logs ... --file-uplink-chunk-size 256`; `ConfigDrivenParser._parse_args` appends these after the config-derived arguments, so they silently override the config file. Verified at this head: `FPRIME_GDS_CONFIG_PATH` pointing at a YAML with `tts-port: 60000` / `tts-addr: 10.0.0.5` yields `arg_ns.tts_port == 50050`, `arg_ns.tts_addr == "0.0.0.0"`; only options whose pytest default is `None` (e.g. `dictionary`) take effect.
Please add a test under `test/fprime_gds/common/testing_fw/` that builds the namespace via `pytest_addoption`/`Parser`, points `FPRIME_GDS_CONFIG_PATH` at a config file, runs the reproduce + `parse_known_args` path, and asserts the config values (including a defaulted option such as `tts-port`) land in the resulting namespace, and that an explicit pytest flag still wins over the file.

src/fprime_gds/executables/cli.py
```diff
@@ -318,14 +321,30 @@ class ConfigDrivenParser(ParserBase):
...
+
+    @classmethod
+    def get_default_configuration(cls):
+        """Get path for (global) default configuration file
+
+        If set, the environment variable (DEFAULT_CONFIGURATION_PATH_ENV) overrides
+        DEFAULT_CONFIGURATION_PATH. Get the path for default configuration file. If unset, will
+        use 'fprime-gds.yml'.
+        """
+        if cls.DEFAULT_CONFIGURATION_PATH_ENV in os.environ:
```
thomas-bc-autowot, line 345: [Test Quality] **could fix** `FPRIME_GDS_CONFIG_PATH` resolution and the new `os.environ.pop` in `set_default_configuration` are untested.

`cli.py` has no unit tests and `test_run_deployment.py::test_as_in_installation_instructions` never sets the variable, so nothing asserts that (a) `--config` defaults to the env path when set, (b) `set_default_configuration()` wins over a previously exported variable, or (c) the fallback to `DEFAULT_CONFIGURATION_PATH` when unset. A small `monkeypatch.setenv` test calling `ConfigDrivenParser.get_default_configuration()` and `ConfigDrivenParser().get_arguments()[("-c", "--config")]["default"]` would cover all three; note `pop` mutates the process environment, so the test should also assert it is removed (or the behavior should be reconsidered).

thomas-bc-autowot, line 345: [Security] **could fix** `general-vulnerability/fail-open-config`: a missing `FPRIME_GDS_CONFIG_PATH` file is silently ignored.

`handle_arguments` only hard-fails when `-c/--config` is in `sys.argv`; an env-selected path that does not exist is skipped without error and every option falls back to built-in defaults (e.g. `--tts-addr 0.0.0.0`, `--file-storage-directory /tmp/<user>`). Before this PR the variable only reached `fprime-gds`; it now applies to every `ConfigDrivenParser` user (pytest fixture, Flask app, CLI tools), so a typo in the deployment environment drops the intended configuration fail-open. Treat an env-provided path like an explicit `--config` in `handle_arguments` (and drop the now-redundant `set_default_configuration` call in `run_deployment.py`, which pops the variable before this check runs):

```python
explicit = (
    "-c" in sys.argv[1:] or "--config" in sys.argv[1:]
    or self.DEFAULT_CONFIGURATION_PATH_ENV in os.environ
)
if explicit and not args.config.exists():
    raise ValueError(f"Specified configuration file '{args.config}' does not exist")
```

src/fprime_gds/executables/cli.py
```diff
@@ -318,14 +321,30 @@ class ConfigDrivenParser(ParserBase):
...
+            return Path(os.environ[cls.DEFAULT_CONFIGURATION_PATH_ENV])
+        return cls.DEFAULT_CONFIGURATION_PATH
```
thomas-bc-autowot, line 347: [Architecture] **could fix** `set_default_configuration` expresses an in-process parser override by mutating the process environment.

Popping `FPRIME_GDS_CONFIG_PATH` from `os.environ` is a hidden global side effect that crosses the launcher's process boundary: `run_deployment.parse_args()` calls this before `launch_comm`/`launch_html`/`launch_plugin` spawn children with the inherited (or `os.environ.copy()`) environment.
Any child that resolves its own configuration through `ConfigDrivenParser` (`flask/app.py`, `gds_app` plugin processes, a nested `fprime_test_api` run) therefore loses the very override this PR introduces and falls back to `fprime-gds.yml`.
Keep the override in the parser's own class state instead:

```suggestion
    _DEFAULT_CONFIGURATION_EXPLICIT = False

    @classmethod
    def set_default_configuration(cls, path: Path):
        """Set path for (global) default configuration file

        Set the path for default configuration file. If unset, will use 'fprime-gds.yml'. Set to None to disable default
        configuration. Calling this function disables the environment variable override.
        """
        cls.DEFAULT_CONFIGURATION_PATH = path
        cls._DEFAULT_CONFIGURATION_EXPLICIT = True

    @classmethod
    def get_default_configuration(cls):
        """Get path for (global) default configuration file

        If set, the environment variable (DEFAULT_CONFIGURATION_PATH_ENV) overrides
        DEFAULT_CONFIGURATION_PATH unless set_default_configuration() has been called. If unset,
        will use 'fprime-gds.yml'.
        """
        if (
            not cls._DEFAULT_CONFIGURATION_EXPLICIT
            and cls.DEFAULT_CONFIGURATION_PATH_ENV in os.environ
        ):
            return Path(os.environ[cls.DEFAULT_CONFIGURATION_PATH_ENV])
        return cls.DEFAULT_CONFIGURATION_PATH
```

src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -106,11 +109,24 @@ def fprime_test_api_session(request):
...
+        [StandardPipelineParser], arguments=reproduced_args, client=True
+    )
```
thomas-bc-autowot, line 124: [Maintainability] **could fix** `maint-clever-code`: `ConfigDrivenParser` still consults `sys.argv` even though `arguments=` is passed here, so this call's behavior is not a function of `reproduced_args`.

`ConfigDrivenParser.handle_arguments` decides "config file was explicitly specified" via `"-c" in sys.argv[1:] or "--config" in sys.argv[1:]` (cli.py ~L491), not via the `arguments` list it was given. Under pytest, `sys.argv` is pytest's own command line and pytest has its own `-c FILE` option; `pytest -c pytest.ini ...` therefore makes this fixture raise "Specified configuration file 'fprime-gds.yml' does not exist" and `sys.exit(-1)` via `ParserBase._parse_args` whenever no `fprime-gds.yml` is in the CWD, even though the comment above says pytest's argv is deliberately kept out of this parse.
A reader of this fixture cannot see that coupling; it is exactly the `sys.argv` leak the comment claims to avoid. Making the explicit-config check operate on the `arguments` actually being parsed (e.g. detect `-c`/`--config` in `arguments` inside `_parse_args` and pass the result through `**kwargs`) removes the hidden input for every caller.

src/fprime_gds/executables/cli.py
```diff
@@ -318,14 +321,30 @@ class ConfigDrivenParser(ParserBase):
 
     DEFAULT_CONFIGURATION_PATH = Path("fprime-gds.yml")
 
+    # Takes precedence over DEFAULT_CONFIGURATION_PATH
+    DEFAULT_CONFIGURATION_PATH_ENV = "FPRIME_GDS_CONFIG_PATH"
```
thomas-bc-autowot, line 325: [Maintainability] **could fix** `maint-code-duplication`: the env-var override now lives in two places — this constant + `get_default_configuration()`, and the pre-existing literal block in `run_deployment.parse_args` (`if "FPRIME_GDS_CONFIG_PATH" in os.environ: ConfigDrivenParser.set_default_configuration(...)`, run_deployment.py L43–48).

That block is now redundant (the parser reads the variable itself) and it hard-codes the string this constant was introduced to own, so renaming the variable here leaves `fprime-gds` reading the old name. Worse, because `set_default_configuration()` now pops the variable, the `fprime-gds` entry point is the one path that deletes `FPRIME_GDS_CONFIG_PATH` from the process environment before `launch_html` copies `os.environ` for the Flask child — every other entry point (pytest fixture, `fprime-cli`, Flask app) keeps it.
Delete the run_deployment.py block (or make it use `ConfigDrivenParser.DEFAULT_CONFIGURATION_PATH_ENV`) so there is one source of truth for the variable name and one place that decides precedence.

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Architecture] **Concur** — also in scope for architectural primitive misuse: the re-injected argparse defaults land in `remaining` after `config_args`, inverting `ConfigDrivenParser`'s documented config-file-then-command-line precedence for every non-`None`-default option.

thomas-bc-autowot, line 122: [Correctness] **could fix** `correctness-unhandled-case`: `ConfigDrivenParser.handle_arguments` decides whether `--config` was given explicitly by scanning `sys.argv[1:]` for `-c`/`--config`; when called from this fixture, `sys.argv` is pytest's own command line.

Trigger: `pytest -c pytest.ini ...` (pytest's `-c FILE, --config-file`) with no `fprime-gds.yml` in the cwd and `FPRIME_GDS_CONFIG_PATH` unset. `handle_arguments` raises `ValueError("Specified configuration file 'fprime-gds.yml' does not exist")`, `ParserBase._parse_args` prints the fprime-gds help and calls `sys.exit(-1)` during fixture setup, and the whole test session errors out. Before this PR the fixture never ran `ConfigDrivenParser`, so this path was unreachable from pytest. Fix: derive "explicit" from the parsed value (e.g. `--config` `default=None`, resolved via `get_default_configuration()` in `handle_arguments` when unset) rather than from `sys.argv`.

src/fprime_gds/executables/cli.py
```diff
@@ -318,14 +321,30 @@ class ConfigDrivenParser(ParserBase):
...
+    @classmethod
+    def get_default_configuration(cls):
+        """Get path for (global) default configuration file
+
+        If set, the environment variable (DEFAULT_CONFIGURATION_PATH_ENV) overrides
+        DEFAULT_CONFIGURATION_PATH. Get the path for default configuration file. If unset, will
+        use 'fprime-gds.yml'.
+        """
+        if cls.DEFAULT_CONFIGURATION_PATH_ENV in os.environ:
+            return Path(os.environ[cls.DEFAULT_CONFIGURATION_PATH_ENV])
```
thomas-bc-autowot, line 346: [Correctness] **suggestion** `correctness-boundary`: an empty `FPRIME_GDS_CONFIG_PATH` is treated as a configured path.

`in os.environ` tests presence, not value, so `export FPRIME_GDS_CONFIG_PATH=` yields `Path("")`, i.e. `Path(".")`. `handle_arguments` then sees `args.config.exists()` true and `open(".")` raises `IsADirectoryError`, which `ParserBase._parse_args` turns into `[ERROR] [Errno 21] Is a directory: '.'` + `sys.exit(-1)` for every `ConfigDrivenParser` user (fprime-gds, Flask app, pytest fixture). Treat an empty value as unset; `run_deployment.py:45` has the same `in os.environ` check and would still pass `Path("")` to `set_default_configuration`, so it needs the same guard.

```suggestion
        if os.environ.get(cls.DEFAULT_CONFIGURATION_PATH_ENV):
            return Path(os.environ[cls.DEFAULT_CONFIGURATION_PATH_ENV])
```

src/fprime_gds/executables/cli.py
```diff
@@ -318,14 +321,30 @@ class ConfigDrivenParser(ParserBase):
...
     @classmethod
     def set_default_configuration(cls, path: Path):
         """Set path for (global) default configuration file
 
         Set the path for default configuration file. If unset, will use 'fprime-gds.yml'. Set to None to disable default
-        configuration.
+        configuration. Calling this function disables the environment variable override.
         """
         cls.DEFAULT_CONFIGURATION_PATH = path
+        os.environ.pop(cls.DEFAULT_CONFIGURATION_PATH_ENV, None)
```
thomas-bc-autowot, line 335: [Correctness] **could fix** `correctness-other`: popping the variable from `os.environ` also removes it from the environment inherited by child processes, which defeats the env-var lookup added in `get_default_configuration` for the Flask GUI.

Trigger: `FPRIME_GDS_CONFIG_PATH=/x/cfg.yml fprime-gds`. `run_deployment.parse_args` (the only in-tree caller) calls `set_default_configuration` exactly when the variable is set, popping it; `launch_html` builds the child env from `os.environ.copy()`, and `flask/app.py` parses `[StandardPipelineParser, ConfigDrivenParser]` with no `--config` in `STANDARD_PIPELINE_ARGUMENTS`, so `--config` falls back to `fprime-gds.yml` in the cwd. The `flask:` section of `/x/cfg.yml` is never applied to the GUI, and a stale `./fprime-gds.yml`, if present, is applied instead. Record the override in a class attribute consulted by `get_default_configuration` instead of mutating `os.environ`, or drop the now-redundant `set_default_configuration` call in `run_deployment.py`.

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Correctness] **Concur** — also in scope for functional correctness: pytest's namespace carries the argparse defaults, so every `store` option with a non-`None` default (`tts-port`, `tts-addr`, `logs`, `log-level-gds`, `zmq-transport`, `file-storage-directory`, `file-uplink-*`) is re-emitted as an explicit argument after `config_args` and overrides the file, contrary to the comment above; registering those options in `pytest_addoption` with `default=None` so only user-supplied flags are reproduced would fix it.

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Maintainability] **Concur** — also in scope for maintainability/readability: a classmethod named `set_default_configuration` silently mutating `os.environ` is a hidden global side effect the call site gives no hint of; recording the override as class state (as suggested) keeps the precedence rule readable in one place (`maint-clever-code`); severity from my scope: **suggestion**.

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Correctness] **Concur** — also in scope for functional correctness: `handle_arguments` classifies an env-selected but missing path as "not specified", so a mistyped `FPRIME_GDS_CONFIG_PATH` silently runs every tool on built-in defaults.

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Maintainability] **Concur** — also in scope for maintainability/readability: the new comment block above ("reproduce only the standard-pipeline flags that pytest actually parsed explicitly") describes behavior `reproduce_cli_args` cannot provide, so whichever mechanism fix lands, the comment must be rewritten to match (`maint-misleading-comment`, **must fix**).

src/fprime_gds/executables/cli.py
thomas-bc-autowot, line 346: [Operational] **suggestion** `ops-config-extreme`: an empty `FPRIME_GDS_CONFIG_PATH` resolves to `Path("")` (the CWD) and aborts every `ConfigDrivenParser` entry point.

`Path("").exists()` is true, so `handle_arguments` prints `[INFO] Reading command-line configuration from: .` and `open(".")` raises `IsADirectoryError` outside its `try`; `ParserBase._parse_args` turns that into `[ERROR] [Errno 21] Is a directory: '.'` and `sys.exit(-1)`. Reproduced at this head with `FPRIME_GDS_CONFIG_PATH=""`.
Reachable from any `export FPRIME_GDS_CONFIG_PATH=` or CI `env:` line whose value expands to empty. Before this PR only `fprime-gds` read the variable; now `fprime-cli`, the Flask GUI and the `fprime_test_api` session fixture (every dependent test ERRORs at setup) fail the same way, with no path in the message to point at the cause.
Treating an empty value as unset removes this mode; a value naming a directory still fails, but with the offending path visible.

```suggestion
        env_path = os.environ.get(cls.DEFAULT_CONFIGURATION_PATH_ENV)
        if env_path:
            return Path(env_path)
```

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Operational] **Concur** — also in scope for operational consequences (`ops-contract-trace`): `reproduce_cli_args` cannot distinguish defaults from operator input, so a `tts-port`/`tts-addr`/`zmq-transport`/`logs` value in the shared config file is honored by `fprime-gds` but silently ignored by the test API, which then connects to a different GDS endpoint than the one the operator configured.

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Operational] **Concur** — also in scope for operational consequences (`ops-failure-path`): reproduced at this head — `pytest -c <ini>` with no `fprime-gds.yml` in the CWD makes this session fixture exit -1, so every test depending on `fprime_test_api` ERRORs at setup; this is a regression for existing `pytest -c` invocations that worked before `ConfigDrivenParser` entered this path; severity from my scope: **must fix**.

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Operational] **Concur** — also in scope for operational consequences (`ops-config-extreme`): a mistyped `FPRIME_GDS_CONFIG_PATH` silently swaps the operator's whole configuration (dictionary, comm selection, endpoint) for built-in defaults with only the absence of the `[INFO] Reading command-line configuration` line as evidence, across every `ConfigDrivenParser` tool this PR now wires to the variable; severity from my scope: **suggestion**.

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Operational] **Concur** — also in scope for operational consequences (`ops-failure-path`): `run_deployment.parse_args` still calls this before `launch_html` copies `os.environ`, so under `fprime-gds` the Flask GUI child resolves `fprime-gds.yml` from the CWD while the launcher used the env-selected file — the two processes can read different `flask:`/`command-line-options` sections with no warning.

src/fprime_gds/common/testing_fw/pytest_integration.py
```diff
@@ -86,9 +88,10 @@ def pytest_configure(config):
...
 
-@pytest.fixture(scope='session')
+
+@pytest.fixture(scope="session")
 def fprime_test_api_session(request):
-    """ Create a session-level fprime test API
+    """Create a session-level fprime test API
 
     This is a pytest session fixture. Using the options added above, this will parse the necessary options for
     connecting the standard pipeline to the running GDS. This pipeline is supplied to the fprime test API returned as
```
thomas-bc-autowot, line 97: [Documentation] **suggestion** `fprime_test_api_session` docstring still says the fixture parses only the pytest-supplied options; it now also reads `fprime-gds.yml` / `$FPRIME_GDS_CONFIG_PATH`.

A `fprime-gds.yml` in the working directory (or the environment variable) now changes how pytest connects, and the plugin exposes no `--config` flag (`pytest_addoption` only mirrors `StandardPipelineParser`), so this docstring is the only in-repo place a user can learn where the file comes from. Wording below assumes the file-then-command-line precedence the PR intends; see the Architecture thread on L118-124 for the current defaulted-option behavior.

```suggestion
    This is a pytest session fixture. Using the options added above, this will parse the necessary options for
    connecting the standard pipeline to the running GDS. Options not given on the pytest command line are read from
    the fprime-gds configuration file ($FPRIME_GDS_CONFIG_PATH, else ./fprime-gds.yml), as `fprime-gds` does; the
    plugin exposes no --config flag, so the file is selected only through that variable or the working directory.
    This pipeline is supplied to the fprime test API returned as
```

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot, line 122: [Documentation] **could fix** `stale-external-doc`: the F Prime user manual documents `fprime_test_api` runs as "start `fprime-gds`, then `pytest ...` with the standard GDS flags" and never mentions `fprime-gds.yml` or `FPRIME_GDS_CONFIG_PATH`; after this change both silently affect pytest runs.

Pages needing a note that a config file in the working directory (or `$FPRIME_GDS_CONFIG_PATH`) now supplies pytest defaults, in `nasa/fprime`: `docs/user-manual/gds/gds-test-api-guide.md` (Quick Start, "run `pytest test/int/test_my_deployment.py`") and `docs/user-manual/gds/reusable-integration-tests.md` (Running the tests: "All other flags ... are the standard GDS pipeline options"). Nothing under `docs/` or `README.md` in this repo describes the config file either, and the PR template's "Documentation Included" entry is blank. No fix is expressible here since the pages live in another repo.

cc @LeStarch @thomas-bc — low-confidence finding, please confirm.

src/fprime_gds/executables/cli.py
```diff
@@ -318,14 +321,30 @@ class ConfigDrivenParser(ParserBase):
 
     DEFAULT_CONFIGURATION_PATH = Path("fprime-gds.yml")
```
thomas-bc-autowot, line 322: [Documentation] **could fix** `ConfigDrivenParser` class docstring (L318-319) says the file is read "(if supplied)" and only states command-line-over-file precedence; the file is now also resolved from `$FPRIME_GDS_CONFIG_PATH` before falling back to `./fprime-gds.yml`, with `-c/--config` beating both.

This docstring is the Sphinx-rendered entry point for the config feature (no other in-repo doc covers it), so the full resolution order `--config` > `$FPRIME_GDS_CONFIG_PATH` > `./fprime-gds.yml`, then command line over file values, belongs there. (Anchored below the offending line; the diff does not include L316-320.)

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Documentation] **Concur** — also in scope for public-API doc comments: the new docstring sentence "disables the environment variable override" does not say it does so by deleting `FPRIME_GDS_CONFIG_PATH` from `os.environ` for the process and its children; whichever design lands, the docstring should state the mechanism.

src/fprime_gds/executables/cli.py
```diff
@@ -1334,11 +1354,11 @@ def get_arguments(self) -> Dict[Tuple[str, ...], Dict[str, Any]]:
                 "type": str,
                 "help": "Set the GUI server address [default: %(default)s]",
             },
-            ("--skip-browser-open",):{
+            ("--skip-browser-open",): {
```
thomas-bc-autowot, line 1357: [Design] **could fix** `design-scope-creep`: roughly half of this diff is formatting-only churn unrelated to the stated intent.

In `cli.py`, 5 of the 8 hunks (blank-line removals, `description`/`dir_name` re-wrapping, this `--skip-browser-open` entry) change no behavior, and `pytest_integration.py` carries docstring/quote/trailing-comma normalization across the whole module. The repository does not enforce a formatter in CI, so this is discretionary churn that obscures the functional change (`ConfigDrivenParser` env lookup + fixture rewiring) and complicates blame/backports. Revert the formatting-only hunks or land them as a separate PR.

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Design] **Concur** — also in scope for design-vs-intent: the stated intent (the fixture picks up the same options as `fprime-gds` from `FPRIME_GDS_CONFIG_PATH`) holds only for options whose pytest default is `None`/`False`; the design should feed `ConfigDrivenParser` only user-supplied flags (e.g. parse `request.config.invocation_params.args` against a defaults-suppressed copy of the pipeline parser) so config-file precedence matches the CLI tools.

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Design] **Concur** — also in scope for code-vs-design: the fixture's own comment defines the design as "pytest's argv never reaches this parse", yet `ConfigDrivenParser.handle_arguments` still keys off `sys.argv`, so `pytest -c <ini>` with no `fprime-gds.yml` in cwd hard-exits the session — a regression for a supported pytest invocation on a code path this PR newly reaches; severity from my scope: **must fix**.

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Design] **Concur** — also in scope for design patterns: a parser setter with a hidden process-environment side effect couples every child process to whichever tool parsed first, and because `run_deployment.parse_args` still calls it the PR's own env-var lookup is defeated for the Flask child; keep the override in class state as suggested and drop the now-redundant `run_deployment.py` block; severity from my scope: **suggestion**.

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Design] **Concur** — also in scope for fault handling: an operator-selected (env) config path that does not exist should fail loudly exactly like `-c/--config` does, not silently degrade every tool to built-in defaults; severity from my scope: **suggestion**.

src/fprime_gds/common/testing_fw/pytest_integration.py
thomas-bc-autowot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime-gds/pull/354#discussion_r4010378865.

src/fprime_gds/executables/cli.py
thomas-bc-autowot: [Summary] **Duplicate** — consolidated into https://github.com/nasa/fprime-gds/pull/354#discussion_r4010375899.

thomas-bc-autowot (changes requested):
## Automated review summary (run 1)

### Per-agent results

| Agent | must fix | suggestion | could fix | future work | outstanding | Verdict |
|---|---:|---:|---:|---:|---:|---|
| Security Vulnerabilities | 0 | 0 | 1 | 0 | 0 | Go |
| Supply Chain / Runner Safety | 0 | 0 | 0 | 0 | 0 | Go |
| F Prime C/C++ Design | 0 | 0 | 0 | 0 | 0 | Go |
| Documentation Currency | 0 | 1 | 3 | 0 | 4 | Go |
| Design | 2 | 2 | 1 | 0 | 5 | No-Go |
| Architecture | 1 | 0 | 1 | 0 | 2 | No-Go |
| Test Quality | 1 | 0 | 1 | 0 | 2 | No-Go |
| Correctness | 1 | 1 | 3 | 0 | 5 | No-Go |
| Operational | 2 | 2 | 1 | 0 | 5 | No-Go |
| Maintainability | 1 | 1 | 2 | 0 | 4 | No-Go |
| **CI safety** | — | — | — | — | — | **Go** |
| **Totals** | 8 | 7 | 11 | 0 | 25 | **No-Go** |

Duplicates consolidated this run: 2 (threads closed by the §5h post-pass; counted once in Totals). Duplicate-close replies are posted, but `resolveReviewThread` was refused for the review token, so both threads remain visibly open.

<details>
<summary>Supply-chain surfaces</summary>

| Surface | Outstanding |
|---|---|
| Dependencies | clean |
| Vendored / submodule | clean |
| Build / test infrastructure | clean |
| Workflows / actions / scripts | clean |
| Generator output | clean |
| Prompt-injection | clean |
| Review-system integrity | clean |

</details>

<details>
<summary>Outstanding must-fix items (2 distinct findings)</summary>

**Test Quality**
- Config-file loading in `fprime_test_api_session` is untested and does not work for defaulted options: `reproduce_cli_args` re-emits argparse defaults after `config_args`, so every non-`None` default (`tts-port`, `tts-addr`, `logs`, …) overrides the config file — also: Architecture, Correctness, Maintainability (misleading comment block), Operational, Design — https://github.com/nasa/fprime-gds/pull/354#discussion_r4010363510

**Maintainability**
- `ConfigDrivenParser.handle_arguments` keys off `sys.argv`, so `pytest -c <ini>` with no `fprime-gds.yml` in the cwd hard-exits the session fixture (regression for a supported pytest invocation); tagged could fix by Maintainability, escalated to must fix by Operational and Design — also: Correctness (duplicate) — https://github.com/nasa/fprime-gds/pull/354#discussion_r4010378865

</details>

### Merge readiness
**Merge readiness: No-Go** — six reviewers report outstanding must-fix items (2 distinct findings: config-file precedence in the `fprime_test_api_session` fixture, and the `sys.argv`-driven `pytest -c` regression).

---

Two focused fixture fixes would put this useful config-propagation change on solid ground.

PR #355, opened 2026-09-14
Harden FileDownlinker.sanitize() against non-native path separators

Imshad18:
## Summary

`FileDownlinker.sanitize()` neutralizes directory separators in the
spacecraft-supplied destination path by replacing `os.sep`. On POSIX this is
correct: `os.sep` is `/`, every separator is removed, and `os.path.join()`
cannot escape the configured downlink directory.

The containment therefore depends on the platform the GDS runs on rather than
on the code. Where `os.sep` is `\`, `/` survives sanitization while the
platform's path layer still honours it as a separator, and an absolute
destination causes `os.path.join()` to discard the base directory entirely:

| destPath | resolved |
|---|---|
| `../../evil.txt` | `C:\evil.txt` |
| `../../../Users/Public/evil.txt` | `C:\Users\Public\evil.txt` |
| `C:/Windows/Temp/evil.txt` | `C:\Windows\Temp\evil.txt` |

`destPath` arrives in the downlink START packet, so it is chosen by the remote
end rather than by the operator. `TransmitFile.open()` adds no further
validation: it calls `Path(filepath).touch(exist_ok=True)` and opens the file
`"rb+"`, so an existing file can be modified at chosen offsets, not merely
created.

## Scope

fprime-gds currently declares only `Operating System :: Unix` / `POSIX`, so
this is **not reachable on a supported platform today** and I am not raising it
as a security issue. This is a defence-in-depth change so the containment
boundary is a property of the code rather than of the host's `os.sep`.

## Change

The destination is reduced to a bare filename. This preserves the existing
"flatten into the downlink directory" behaviour while removing every separator,
drive prefix, and NTFS alternate-data-stream marker.

## Tests

Adds a parametrized suite covering 19 destination paths. Each case asserts both
the expected sanitized name and that `os.path.join()` of that name stays inside
the downlink directory under **both** `posixpath` and `ntpath`, so the
guarantee no longer depends on which platform the tests run on.

Verified locally: 38/38 new checks pass, and the pre-existing suite is
unchanged (the same 2 failures / 6 errors occur on a clean `devel` checkout in
this environment and are unrelated to this change).


thomas-bc:
No linked issue, PR template is not used. Closing.

PR #356, opened 2026-09-14
Harden FileDownlinker.sanitize() against non-native path separators

Imshad18:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#5963 |
|**_Documentation Included (y/n)_**| n (docstring updated; no user-facing doc change) |

---
## Change Description

`FileDownlinker.sanitize()` reduces the downlink destination to a bare filename instead of
replacing only `os.sep`:

```python
flattened = filename.replace("\\", "/")
base = posixpath.basename(posixpath.normpath(flattened))
base = base.replace(":", "_")
if base.strip(".") == "":
    return "_unnamed_downlink"
return base
```

Both `/` and `\` are treated as separators regardless of the host's `os.sep`; `:` is
replaced so it cannot reintroduce a drive specifier or an NTFS alternate data stream; and a
name consisting only of dots is replaced, since `os.path.join()` would otherwise resolve it
to the downlink directory itself or its parent.

## Rationale

Detail is in nasa/fprime#5963. In short: the current `filename.replace(os.sep, "_")` makes
the containment guarantee a property of the host platform rather than of the code. On POSIX
it is correct. Where `os.sep` is `\`, `/` survives sanitization while the path layer still
honours it, and an absolute destination causes `os.path.join()` to drop the base directory.

`destPath` comes from the downlink START packet, so it is chosen by the remote end, and
`TransmitFile.open()` adds no further validation (`Path(filepath).touch(exist_ok=True)` then
`open(..., "rb+")`).

fprime-gds declares only Unix/POSIX support, so **this is not reachable on a supported
platform today and is not being reported as a vulnerability.** This is a defence-in-depth
change so the boundary holds regardless of host, and does not silently weaken if platform
support ever widens. Behaviour on POSIX is unchanged: destinations still flatten into the
downlink directory.

Adds `test/fprime_gds/common/files/test_downlinker.py`: 19 destination paths, each asserted
twice — the expected sanitized name, and that `os.path.join()` of that name stays inside the
downlink directory under **both** `posixpath` and `ntpath`. The containment assertion is
made against both path flavours deliberately, so the guarantee no longer depends on which
platform CI runs on.

Verified locally: 38/38 new checks pass. The pre-existing suite is unchanged — 596 pass, and
the same 2 failures / 6 errors reproduce on a clean `devel` checkout in this environment
(WSL permission issues in `test_plugins.py` / `test_run_deployment.py`), so they are
unrelated to this change.

Worth reviewing: whether `_unnamed_downlink` is the preferred fallback name, and whether
flattening (rather than rejecting) a path-bearing destination is still the behaviour the
team wants.

## Future Work

None required for this change. If Windows ever becomes a supported platform for the GDS,
`TransmitFile.open()` and the `__log_dir` join in the same module would be worth a similar
pass.

---

Supersedes #355, which I opened without a linked issue or the PR template — my mistake.


thomas-bc:
See https://github.com/nasa/fprime/issues/5963

